PR #6, opened 2015-04-27
get latest version with 'go get'

cryptix:
Without the `-u` flag people won't get the latest version if they fetched it before.


jbenet:
argh yeah. thanks

PR #10, opened 2015-12-14
Added a downgrading guide doc

RichardLitt:
Thought this might be useful: I had to ask @diasdavid how to do this.

### DIFF
--- a/downgrading.md
+++ b/downgrading.md
@@ -1,0 +1,38 @@
+## Automatically using `ipfs-update`
+
+If you want to do a downgrade (and not a clean install of an older version) you can revert your repo using the fs-migration tools:
+
+```sh
+$ ipfs-2-to-3 -revert -path=$IPFS_PATH
+```
+
+That should be it. You're all set.
+
+## Manually downgrading IPFS versions
+
+**You should not have to do this manually. Automatically downgrading is the preferred method.**
+
+Manually downgrading your IPFS version (for instance, from `0.4.0` to `0.3.8`) - which would involve keeping your data - is currently **not possible**. You can't use the same repository simultaneously for both versions. Here is how you switch. However, you can reinstall from scratch.
+
+First, go to your repo for IPFS, check out the right branch, and re-install.
+
+```sh
+cd $GOPATH/src/github.com/ipfs/go-ipfs
+git checkout master # (or the version you want)
+go install ./cmd/ipfs
+```
+
+Then to backup your repo:
+
+```sh
+mv ~/.ipfs ~/.ipfs.bak
+```
+
+Finally, run:
+
+```sh
+ipfs init
+ipfs daemon
+```
+
+And you should be all set.
--- a/downgrading.md
+++ b/downgrading.md
@@ -19,7 +19,7 @@ First, go to your repo for IPFS, check out the right branch, and re-install.
 ```sh
 cd $GOPATH/src/github.com/ipfs/go-ipfs
 git checkout master # (or the version you want)
-go install ./cmd/ipfs
+make install
 ```
 
 Then to backup your repo:


downgrading.md
```diff
@@ -0,0 +1,26 @@
+## Downgrading IPFS versions
+
+Downgrading your IPFS version (for instance, from 4.0 to 3.8) is not as complicated as upgrading. You can't use the same repository simultaneously for both versions. Here is how you switch
```
jbenet: `0.4.0` and `0.3.8`

downgrading.md
```diff
@@ -0,0 +1,26 @@
...
+
+First, go to your repo for IPFS, check out the right branch, and re-install.
```
jbenet: let's put this whole section under:

```
## Manual Way
```
Because https://github.com/ipfs/ipfs-update has landed and it lets you downgrade too.

jbenet: and yeah we should add a section above:

```
## With `ipfs-update`
```

RichardLitt:
Added a section for using ipfs-update and a TODO; I can write that, but will need help as I'm unfamiliar with ipfs-update.

jbenet:
Cc @whyrusleeping

whyrusleeping:
I would add a bit on what the actions taken in your code actually do, your new daemon won't have any of the same data as before.

If you really want to do a downgrade (and not a clean install of an older version) you would have to revert your repo:
```
$ ipfs-2-to-3 -revert -path=$IPFS_PATH
```

jbenet:
> your new daemon won't have any of the same data as before.

no guide that we give users should ever do this (unless it makes it absolutely clear with a huge warning). i propose we stick to always getting users to downgrade the migration.

 we can mention that "you can just get rid of your old repo, or move it to a backup location and init a new node" as a _secondary_ option.

diasdavid:
Oh, there is a `-revert` option? nice!

jbenet:
> Oh, there is a -revert option? nice!

:)

RichardLitt:
> ipfs-2-to-3 -revert -path=$IPFS_PATH

Does this work for ifps-3-to-4?

jbenet:
yeah they should all be the same. though 3-to-4 isnt merged yet

RichardLitt:
I _think_ this is the edit suggested, but I am not sure. I've only ever used the method described in the manual section, because that was what was suggested for me. Thoughts?

whyrusleeping:
I'll admit the numbering is a tad confusing '2' here refers to ipfs 0.3.*, and '3' refers to ipfs 0.4.0

jbenet:
The numbers are independent sequences. The fs-repo versions 2 and 3
coincide with go-ipfs versions 0.3 and 0.4, but are different numbers.

RichardLitt:
So, `ipfs-2-to-3` doesn't actually refer to IPFS 2, or 3? That is really confusing. How about `ipfs-fs-2-to-3` or something as a name, instead?

whyrusleeping:
i mean, thats why its the 'fs-repo-migrations', but yeah. maybe we could rename slightly

lgierth:
`fs-repo-2-to-3` would match `fs-repo-migrations`. Also just in case: blue! :):)

jbenet:
+1 for fs-repo-2-to-3

RichardLitt:
Is that all needs to happen to make that name change a change of the dir_names in this directly from `ipfs-*` to `fs-*`?

downgrading.md
```diff
@@ -0,0 +1,38 @@
...
+
+```sh
+$ ipfs-2-to-3 -revert -path=$IPFS_PATH
+```
+
+That should be it. You're all set.
+
+## Manually downgrading IPFS versions
+
+**You should not have to do this manually. Automatically downgrading is the preferred method.*
```
Dignifiedquire: One `*` either too much or too little

RichardLitt:
Good catch @Dignifiedquire, thanks.

chriscool:
@whyrusleeping or @jbenet could you take a look at merging this? Thanks!

downgrading.md
```diff
@@ -0,0 +1,38 @@
...
+$ ipfs-2-to-3 -revert -path=$IPFS_PATH
+```
+
+That should be it. You're all set.
+
+## Manually downgrading IPFS versions
+
+**You should not have to do this manually. Automatically downgrading is the preferred method.**
+
+Manually downgrading your IPFS version (for instance, from `0.4.0` to `0.3.8`) is not as complicated as upgrading. You can't use the same repository simultaneously for both versions. Here is how you switch. Note that this is actually a clean install, and that your daemon won't have the same data as before.
```
jbenet: No, this is not downgrading, this is installing a different version from scratch. 

Downgrading the version of a repo MUST ALWAYS preserve all user data.

RichardLitt: I feel like we make that clear? But I've pushed a slightly different version making this blatant.

chriscool:
Ping!

whyrusleeping:
This should be updated, the instructions it gives are incorrect, it *is* possible to downgrade safely without losing data using the `ipfs-X-to-Y` tools. We should document that here.

RichardLitt:
@whyrusleeping I am confused. It looks like it says that, to me. That's the first section.

RichardLitt:
@whyrusleeping What do you need from me, here?

downgrading.md
```diff
@@ -0,0 +1,38 @@
...
+**You should not have to do this manually. Automatically downgrading is the preferred method.**
+
+Manually downgrading your IPFS version (for instance, from `0.4.0` to `0.3.8`) - which would involve keeping your data - is currently **not possible**. You can't use the same repository simultaneously for both versions. Here is how you switch. However, you can reinstall from scratch.
+
+First, go to your repo for IPFS, check out the right branch, and re-install.
+
+```sh
+cd $GOPATH/src/github.com/ipfs/go-ipfs
+git checkout master # (or the version you want)
+go install ./cmd/ipfs
```
lgierth: this should be `make install` now

RichardLitt:
@lgierth Fixed!

downgrading.md
```diff
@@ -17,9 +17,7 @@ Manually downgrading your IPFS version (for instance, from `0.4.0` to `0.3.8`) -
 First, go to your repo for IPFS, check out the right branch, and re-install.
 
 ```sh
-cd $GOPATH/src/github.com/ipfs/go-ipfs
-git checkout master # (or the version you want)
-go install ./cmd/ipfs
+make install
```
lgierth: Oh I meant only that `go install` is `make install` now :)

RichardLitt: Fixed again.

hsanjuan:
This is a bit old, and this repo is probably not the place to document a general downgrade process now we have dedicated websites.

RichardLitt:
Sounds good. Thanks, @hsanjuan.

Issue #16, opened 2016-01-07
Migrating downward to a previous version is not documented


chriscool:
It is not clear from the doc and from running `fs-repo-migrations -h` if it is possible to go back to a previous version or not. And if it is possible it is not clear how.

RichardLitt:
I've been working on this in https://github.com/ipfs/fs-repo-migrations/pull/10.

chriscool:
@RichardLitt thanks!

hsanjuan:
See linked issue in linked issue

Issue #12, opened 2015-12-30
fs-repo-migrations should have a way to show its own version


chriscool:
Or at least it should have a way to tell which migrations it can handle, without trying the migration.
Otherwise it is difficult for people to know if they have the right version of fs-repo-migrations for their needs.

chriscool:
Ok maybe `-v` is to show the highest version the tool can handle, but it is not clear from the help message "print highest repo version and exit".

"highest repo version" could mean "highest repo version found on the computer" at least that's what I thought first when reading it.

Issue #11, opened 2015-12-30
fs-repo-migrations should not ignore arguments


chriscool:
Ignoring arguments can lead people to wrongly think that it has some hidden features:

```
> fs-repo-migrations versions
Found fs-repo version 2 at /home/utilisateur/.ipfs
Do you want to upgrade this to version 3? [y/n] n

> fs-repo-migrations sdff sdfsf sdfsdf 
Found fs-repo version 2 at /home/utilisateur/.ipfs
Do you want to upgrade this to version 3? [y/n] n
```

PR #17, opened 2016-01-07
sharness: add t0040-migration-2-3.sh

chriscool:
This is a first test to fully test migration from 2 to 3.

It would help if someone could have a look at issues: #16 #12 and #11 especially #16.

cc @whyrusleeping


chriscool:
About #16 @RichardLitt already answered sorry!

whyrusleeping:
LGTM

Issue #18, opened 2016-01-09
Table showing repo version corresponding to ipfs versions?


chriscool:
It would be nice if there was a table showing which repo version corresponds to which ipfs versions.
Also if fs-repo-migration gets its own version number (see issue #12), these version numbers could be added to this table too.

whyrusleeping:
repo version '1' belongs to a version that I think you can only build using go pre1.4. repo version '2' is everything below 0.4.0, and version '3' is version 0.4.0 and on.

jbenet:
> It would be nice if there was a table showing which repo version corresponds to which ipfs versions.

agreed.

table:
```
version 1: 0.0.0 - 0.2.3
version 2: 0.3.0 - 0.3.11
version 3: 0.4.0 - current
```
(see https://github.com/ipfs/go-ipfs/blob/master/CHANGELOG.md#030---2015-04-20 for version 2)

chriscool:
Fixed by PR #19.

PR #19, opened 2016-01-10
README: add a 'When should I migrate' section

chriscool:
This is to fix issue #18 (Table showing repo version corresponding to ipfs versions?)


RichardLitt:
Weren't we going to rename fs-repo-migration numbers, so that this confusion doesn't happen?

RichardLitt:
Otherwise, given #18, this LGTM

Issue #21, opened 2016-02-15
make the 2-3-pins test check if objects are there too


jbenet:
the 2-3pins test checks the pins but does not:
- check the objects are there
- run gc 
between every change of version

it also does not remove pins objects. 

an important goal of these tests is to exercise the pinset across versions, including deletion of pins and GCs that do remove some content and leave other content.

jbenet:
this migration is really critical to get right. these tests should build on this: https://github.com/ipfs/go-ipfs/issues/2343

As things are now, I'm far from certain this migration will proceed without any hitch for users. ipfs MUST NEVER lose any data, and i'm not confident in the tests yet (either for go-ipfs@0.4.0 nor fs-repo-migrations 2-3). this is going to be the first major data migration for users and we need to be at the top of our certainty.

whyrusleeping:
Hey @chriscool could you add these checks to the sharness tests here?

chriscool:
Yeah, I will take a look at adding something like that.
I wanted to add tests using the daemon first, then I looked at iptb to launch the daemon, then it appeared that iptb was not using gx yet...

whyrusleeping:
@chriscool in the meantime we can use go get to pull down iptb (`go get github.com/whyrusleeping/iptb`) it has very few external dependencies.

chriscool:
Yeah I will have another look at using iptb.

chriscool:
So we don't use iptb but we do check the objects are there and run gc between every change of version.

PR #22, opened 2016-02-17
Check objects and add gc

chriscool:
This should take care of the 2 main points in issue #21 (make the 2-3-pins test check if objects are there too).

### DIFF
--- a/sharness/t0050-migration-2-3-pins.sh
+++ b/sharness/t0050-migration-2-3-pins.sh
@@ -102,6 +102,32 @@ test_expect_success "list of pinned object hasn't changed" '
 	test_sort_cmp all_actual1 all_actual2
 '
 
+test_expect_success "objects are still there" '
+	exec_docker "$DOCID" "ipfs cat $HASH_FILE6" >FILE6_b &&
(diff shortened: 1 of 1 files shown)


chriscool:
There is the following error when calling `ipfs repo gc` after the migration:

```
expecting success: 
	exec_docker "$DOCID" "ipfs repo gc"

Error: merkledag: not found

not ok 23 - 'ipfs repo gc' succeeds
```

@whyrusleeping could you take a look?

chriscool:
It looks like the "Error: merkledag: not found" comes from: https://github.com/ipfs/go-ipfs/blob/master/merkledag/merkledag.go#L102

whyrusleeping:
@chriscool rebase on latest master and give it another try, I also fixed an issue reverting from 3 to 2 here: https://github.com/ipfs/fs-repo-migrations/pull/23

chriscool:
@whyrusleeping  rebase done. Yeah, at least the tests pass on my machine. Thanks!

chriscool:
Travis is happy too :-)

whyrusleeping:
@chriscool awesome, thanks!

PR #23, opened 2016-02-18
fix revert 3 to 2 code

whyrusleeping:
We werent using the right pinner to set the revert up


whyrusleeping:
@lgierth said it fixed his issues, tests are green. :ship:

PR #28, opened 2016-03-18
Improve migration stress tests

chriscool:
This makes it possible to check if the ipfs daemon started properly.


whyrusleeping:
thanks @chriscool !

PR #29, opened 2016-03-19
Migration stress

whyrusleeping:
This adds a stress test for the 2 to 3 migration

### DIFF
--- a/sharness/t0060-migration-stress.sh
+++ b/sharness/t0060-migration-stress.sh
@@ -34,23 +34,12 @@ test_docker_wait_for_file() {
 
 test_install_version "v0.3.11"
 
-test_expect_success "'ipfs init' succeeds" '
-	export IPFS_PATH=/root/.ipfs &&
-	exec_docker "$DOCID" "IPFS_PATH=$IPFS_PATH BITS=2048 ipfs init" >actual 2>&1 ||
-	test_fsh cat actual
-'
+test_init_daemon "$DOCID"
 
-test_expect_success "'ipfs daemon' succeeds" '
-	exec_docker "$DOCID" "ipfs daemon >actual_daemon 2>daemon_err &"
-'
-
-test_expect_success "api file shows up" '
-	test_docker_wait_for_file "$DOCID" 20 100ms "$IPFS_PATH/api"
-'
+test_start_daemon "$DOCID"
 
 test_expect_success "make a couple files" '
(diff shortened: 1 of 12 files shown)


chriscool:
"No output has been received in the last 10m0s..." Travis test failures.

whyrusleeping:
heh, probably because these tests take longer than 10 minutes to run. especially on whatever slow disks travis vms use.

What would be the easy way to turn down these numbers in CI?

chriscool:
Yeah, maybe adding a lot of random files takes too long. In this case the obvious way to turn down these numbers is to put smaller ones in the following command:

```
    drun "$GUEST_RANDOM_FILES -depth=6 -dirs=7 -files=10 manyfiles" > filenames
```

whyrusleeping:
yeah, the amount of files i'm adding is exactly how much i want. I just think we should turn it down if CI is on. It looks like theres a variable `TRAVIS` that gets set when travis runs things, so we can check that and turn things down

sharness/t0060-migration-stress.sh
```diff
@@ -4,6 +4,15 @@ test_description="Test migration 2 to 3 with lots of objects"
 
 . lib/test-lib.sh
 
+# setup vars for tests
+DEPTH=6
+PINTOTAL=2000
+if [ ! -z "$CI" ]; then
```
chriscool: Maybe `if test -n "$CI" ...`

jbenet: Travis can run longer than 10m, just needs to receive output from the test,
the check is based on "10m since last output received"

chriscool: Maybe we could try piping `ipfs add` into `tee`, so that its output still goes to stdout.
We will not detect if `ipfs add` fails, but that's probably ok.

chriscool:
I sent a commit that uses 'tee'. Feel free to remove it if you don't like it.

whyrusleeping:
@chriscool LGTM. And the tests work nicely. I think this is ready to merge, do you?

chriscool:
Yeah, you can merge this, though perhaps with my "t0060: pipe 'ipfs add' into 'tee' for Travis" commit your "lower stress numbers for CI" commit is not necessary anymore.
Or we could use an EXPENSIVE prereq like in go-ipfs to have tests with high stress numbers, but this can be done in another PR.

whyrusleeping:
@chriscool agreed. feel free to PR any other changes you feel necessary, i'll go ahead and merge as is

PR #35, opened 2016-03-24
More migration stress

chriscool:
Add revert tests to t0060-migration-stress.sh and also add an EXPENSIVE prereq to make them faster in the normal case.

### DIFF
--- a/sharness/t0060-migration-stress.sh
+++ b/sharness/t0060-migration-stress.sh
@@ -5,11 +5,14 @@ test_description="Test migration 2 to 3 with lots of objects"
 . lib/test-lib.sh
 
 # setup vars for tests
-DEPTH=6
-PINTOTAL=2000
-if [ ! -z "$CI" ]; then
-	DEPTH=3
-	PINTOTAL=200
+
+DEPTH=3
+PINTOTAL=200
+
+if test_have_prereq EXPENSIVE
+then
+	DEPTH=6
+	PINTOTAL=2000
 fi
 
 PINEACH=$(expr $PINTOTAL / 2)
--- a/sharness/t0060-migration-stress.sh
+++ b/sharness/t0060-migration-stress.sh
@@ -7,15 +7,25 @@ test_description="Test migration 2 to 3 with lots of objects"
 # setup vars for tests
 
 DEPTH=3
-PINTOTAL=200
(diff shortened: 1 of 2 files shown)


chriscool:
There is this failure:

```
expecting success: 

	comm -23 start_refs after_revert_refs > missing_refs &&
	test_cmp missing_refs empty_refs_file

> diff -u missing_refs empty_refs_file
--- missing_refs	2016-03-24 09:38:36.927208538 +0000
+++ empty_refs_file	2016-03-24 09:38:32.932312062 +0000
@@ -1 +0,0 @@
-QmemK5Vqg4Cao5otFSebPm5CBEHfuQmE2KLbxHGA11Gm3p

not ok 48 - refs look right
```

@whyrusleeping could you take a look at it. It seems to me that it might be a bug.

whyrusleeping:
@chriscool this seems normal. The `start_refs` file contains refs before the initial migration. After the migration occurs, we run a GC. which means we drop refs that arent pinned.

What we should do here is, after the tests that i ran, but before the revert we should add a bunch more random content (and maybe a few pins too), collect the refs into a new `start_refs` file, and then run the test after the revert.

chriscool:
@whyrusleeping ok I just pushed an amended commit with the changes you suggest and it works on my machine!

sharness/t0060-migration-stress.sh
```diff
@@ -179,6 +193,91 @@ test_expect_success "no pinned objects are missing from local refs" '
...
+test_expect_success "'ipfs-2-to-3 -revert' succeeds" '
+	drun "$GUEST_IPFS_2_TO_3 -revert -path=/root/.ipfs" >actual
+'
+
+test_expect_success "'ipfs-2-to-3 -revert' output looks good" '
+	grep "writing keys:" actual ||
+	test_fsh cat actual
+'
+
+test_install_version "v0.3.10"
```
whyrusleeping: should probably use `v0.3.11` here. since its the latest version

chriscool: Ok I will change that.

chriscool:
Now using v0.3.11 everywhere.

whyrusleeping:
Cool, this looks great to me!

Issue #41, opened 2016-07-02
Trying out the repo migrations from 3 to 4


daviddias:
I've managed to successfully migrate my repo_v3 to a repo_v4, but I was unable to revert it back. It seems that there is no way to revert at the high level tool

```
» fs-repo-migrations -h
Usage of fs-repo-migrations:
  -to int
        specify version to upgrade to (default 4)
  -v    print highest repo version and exit
  -y    answer yes to all prompts
```

And using the dedicated binary doesn't work

```
 ⭠ master± ⮀ ~/code/go-projects/src/github.com/ipfs/fs-repo-migrations/ipfs-3-to-4 ⮀
» ./ipfs-3-to-4 --revert
Usage of ./ipfs-3-to-4:
  -f    whether to force a migration (ignores warnings)
  -path string
        file path to migrate for fs based migrations
  -revert
        whether to apply the migration backwards
  -verbose
        enable verbose logging
 ⭠ master± ⮀ ~/code/go-projects/src/github.com/ipfs/fs-repo-migrations/ipfs-3-to-4 ⮀
» ./ipfs-3-to-4 -revert
Usage of ./ipfs-3-to-4:
  -f    whether to force a migration (ignores warnings)
  -path string
        file path to migrate for fs based migrations
  -revert
        whether to apply the migration backwards
  -verbose
        enable verbose logging
```

I've also found it to be weird to have full word flags behind a `-` and not two, like `--verbose` instead of `-verbose`.

As a go-ipfs user, I was expecting for the installation process to be a bit more verbose or even use gx, currently all that is printed is:

```
» make install
go install
```

It would be useful to have a message saying that it was installed and that users can type fs-repo-migrations on their terminal.

whyrusleeping:
@diasdavid I updated the `ipfs-3-to-4` tool to say that the `path` flag is required. I also added a note to the makefile as you said.

Thanks for the feedback :)

jbenet:
can these observations be turned into sharness tests?

chriscool:
@jbenet I am taking a look at that...

chriscool:
Actually when you run `ipfs-3-to-4 -revert` you get:

```
> ipfs-3-to-4 -revert
Usage of ../bin/ipfs-3-to-4:
  -f    whether to force a migration (ignores warnings)
  -path string
        file path to migrate for fs based migrations (required)
  -revert
        whether to apply the migration backwards
  -verbose
        enable verbose logging
```

[...]

PR #44, opened 2016-07-13
Migration without path

chriscool:
This is to fix some bits described in issue #41


chriscool:
The Travis failure is because the ipfs daemon could not be killed.
It doesn't seem related.

whyrusleeping:
Thanks @chriscool !!

PR #47, opened 2017-01-19
migration from 4 to 5

whyrusleeping:
Still WIP, waiting on flatfs code to be finalized


ipfs-4-to-5/migration/migration.go
```diff
@@ -0,0 +1,211 @@
...
+	if err := flatfs.Create(tempffs, flatfs.NextToLast(2)); err != nil {
+		if err2 := revertStep2(ffspath); err2 != nil {
+			log.Error(err2)
+		}
+		return err
+	}
+
+	revert3 := func(mainerr error) error {
+		log.Error("failed to convert flatfs datastore: %s", mainerr)
+		log.Log("attempting to revert...")
```
kevina: This might not be the best option, espacally if the process fails halfway though and can easily be corrected.  It way be better to give to allow the conversion to be restartable (you would have to skip the initial steps) and just leave things as is so the user can fix the problem and try again.

That being said.  I'm okay with leaving things as is in the interest of getting something out quickly.

kevina:
Note if users run into problems is tricky situationist I will be happy to walk them thorough fixing the problem.  Most of these steps can be done by hand using the "flatfs" utility.

ipfs-4-to-5/migration/migration.go
whyrusleeping: Nope, the migrate process *must* either succeed, or leave the repo untouched.

whyrusleeping:
@kevina Yeah, having that tool around will be nice. Do you mind writing up a little bit on potential problems people might encounter and how to use the tool to address them?

ipfs-4-to-5/migration/migration.go
kevina: I'm not sure that is consistent with what was done previously but okay.

Note, that the revert could fail.  If I knew this is what you wanted to do I might of considered using hardlinks instead of moving, but I'm not sure how well that will work on windows.

kevina:
> @kevina Yeah, having that tool around will be nice. Do you mind writing up a little bit on potential problems people might encounter and how to use the tool to address them?

Okay, but I think a better approach would be to have a restartable migration, that will just stop on error, let the user fix the problem and then restart.  It will also be more robust in the sense that it can be interrupted with ill side effects. 

The `flatfs` util won't be able to do a full upgrade as some, for example it won't update the version number of the repo as that is outside the datastore.

kevina:
Note: The "restartable migration" doesn't need to be the default operation, but can be provided as an option to handle tricky situations or to recover from an interrupted migration.

I will be happy to provide this in a separate p.r.  It doesn't have to block the release of 0.4.5.

ipfs-4-to-5/migration/migration.go
jbenet: - +1 on migration proc MUST either succeed or roll back. think of atomic transactions.
   - if revert fails, that's when we fully halt and ask the user to fix.
- i also considered hardlinks. it has issues.

whyrusleeping:
I would be fine having that as a passed in option.

kevina:
Okay I will do it as a separate p.r.  Hopefully by the end of tomorrow.

ipfs-4-to-5/migration/migration.go
```diff
@@ -0,0 +1,224 @@
...
+
+	revert1 := func(e error) error {
+		err := os.Rename(ffspath, basepath)
+		if err != nil {
+			log.Error(err)
+		}
+		return e
+	}
+
+	log.Log("> Upgrading datastore format to have sharding specification file")
```
kevina: Is there any special reason some of the log entries start with `>`?  Does the `>` do something special?

whyrusleeping: No, just there to differentiate the main steps from other log messages

whyrusleeping:
@kevina tests are failing now for some reason...

ipfs-4-to-5/migration/migration.go
kevina: okay, just wondering

PR #49, opened 2017-01-23
4-to-5 migration improvments

kevina:
Try to recover from an interrupted migration in `ipfs-4-to-5` and add a `-no-revert` option to avoid automatically reverting in case of failure.

This doesn't cover all failure cases but covers the most common ones (mainly flatfs.Move being interrupted or failing for some reason).  For other failure cases it should be possible to manually fix up the repo and then rerunning `ipfs-4-to-5`.

### DIFF
--- a/go-migrate/migrate.go
+++ b/go-migrate/migrate.go
@@ -29,6 +29,13 @@ type Migration interface {
 	Revert(Options) error
 }
 
+type SupportNoRevert interface {
+	Migration
+
+	// Supports the NoRevert option
+	SupportNoRevert() bool
+}
+
 func SplitVersion(s string) (from int, to int) {
 	_, err := fmt.Scanf(s, "%d-to-%d", &from, &to)
 	if err != nil {
--- a/go-migrate/migrate.go
+++ b/go-migrate/migrate.go
@@ -29,13 +29,6 @@ type Migration interface {
 	Revert(Options) error
 }
 
-type SupportNoRevert interface {
-	Migration
-
-	// Supports the NoRevert option
-	SupportNoRevert() bool
-}
-
 func SplitVersion(s string) (from int, to int) {
 	_, err := fmt.Scanf(s, "%d-to-%d", &from, &to)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


kevina:
not sure how this will work,  I made the pull request on top of another pull request, as in it is based on ` feat/4-to-5 ` and not master

ipfs-4-to-5/migration/migration.go
```diff
@@ -53,7 +53,16 @@ func (m Migration) Apply(opts migrate.Options) error {
 	basepath := filepath.Join(opts.Path, "blocks")
 	ffspath := filepath.Join(opts.Path, "blocks-v4")
 	if err := os.Rename(basepath, ffspath); err != nil {
-		return err
+		if os.IsNotExist(err) {
+			fi, err2 := os.Stat(basepath)
```
whyrusleeping: don't you mean to stat ffspath here?

kevina: Yep.

kevina:
@whyrusleeping okay.  I am going to add a "--no-revert" option as part of the p.r. this will make failure and recovery easier to test.

kevina:
@whyrusleeping assuming the tests pass this should be ready now

kevina:
note that if you "merge" this it should go into #47 and not master ...

go-migrate/migrate.go
```diff
@@ -29,6 +29,13 @@ type Migration interface {
 	Revert(Options) error
 }
 
+type SupportNoRevert interface {
```
whyrusleeping: This is weird... I would just make it part of the interface and backfill it

whyrusleeping: Or make it more general somehow. This is a very weird single purpose interface

kevina: @whyrusleeping I only did this becuase the readme says:

> Frozen. After the tool is written, all code must be frozen and vendored.

I took that that old versions of the conversion tool should not change.  I will happy to backfill it if you think it is better.

whyrusleeping: Ah, hrm...  Maybe then a simple array of booleans for checking which versions are supported would be the cleanest solution here

kevina: okay will do

PR #53, opened 2017-02-15
update to fix issue with extraneous files in flatfs during migration

whyrusleeping:


whyrusleeping:
As a side note, i'd love to switch to gx vendoring here...

Kubuxu:
Yeah, but is gx ready for multiple sub-packages? It should be, we could have just per migration package.json.

whyrusleeping:
@Kubuxu Yeah, it should work for subpackages just fine

PR #55, opened 2017-02-16
fs-repo-migrations can now revert

jbenet:
This PR enables fs-repo-migrations to run migrations backwards (reverts). I am not sure how it will play with go-ipfs or if it will cause invariants to be violated.

```
> cat ~/.ipfs/version
5

> ./fs-repo-migrations -to 4 -y
Found fs-repo version 5 at /Users/jbenet/.ipfs
===> Running migration 5 to 4...
reverting migration
Moving Keys...
90600 keys so far
Cleaning Up...
All Done.
lowered version number to 4
===> Migration 5 to 4 succeeded!

> cat ~/.ipfs/version
4

> ./fs-repo-migrations -to 5 -y
Found fs-repo version 4 at /Users/jbenet/.ipfs
===> Running migration 4 to 5...
applying 4-to-5 repo migration
locking repo at "/Users/jbenet/.ipfs"
  - verifying version is '4'
> Upgrading datastore format to have sharding specification file
> creating a new flatfs datastore with new format
> converting current flatfs datastore to new format
Moving Keys...
90610 keys so far
Cleaning Up...
All Done.
> moving new datastore into place
> moving transferred datastore back into place
updated version file
===> Migration 4 to 5 succeeded!

> cat ~/.ipfs/version
5
```

PR #56, opened 2017-02-16
Reversibility tests

jbenet:
these test the code in https://github.com/ipfs/fs-repo-migrations/pull/55

### DIFF
--- a/main.go
+++ b/main.go
@@ -61,8 +61,8 @@ func GetIpfsDir() (string, error) {
 	return "", err
 }
 
-func runMigration(n int) error {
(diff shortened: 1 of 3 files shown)


jbenet:
Im not sure these work, so putting in a different PR than https://github.com/ipfs/fs-repo-migrations/pull/55 -- sorry, cant fix these up.

magik6k:
There is revert mode, though not all migrations test it

Stebalien:
Status: `ipfs-update` doesn't pass the `--revert-ok` flag so the migration backwards correctly fails. We'll likely need to add a flag to `ipfs-update` to allow this.

PR #62, opened 2017-06-19
repo migration from 5 to 6

kevina:
see https://github.com/ipfs/go-ipfs/pull/3575

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
   - TEST_VERBOSE=1
 
 script:
-  - make test
+  - make -k test
 
 # For docker containers
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,7 @@ env:
   - TEST_VERBOSE=1
 
 script:
-  - make -C sharness
-  - go test ./ipfs-5-to-6/migration/
+  - make test
 
 # For docker containers
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,8 @@ env:
   - TEST_VERBOSE=1
 
 script:
-  - cd sharness
-  - make
+  - make -C sharness
+  - go test ./ipfs-5-to-6/migration/
 
 # For docker containers
 
--- a/ipfs-5-to-6/migration/ci_config.go
+++ b/ipfs-5-to-6/migration/ci_config.go
@@ -1,0 +1,57 @@
+package mg5
+
+import (
+	"strings"
+)
+
+// ciConfig is a case insensitive map to represent a IPFS
+// configuration when unmarshaled as a map[string]interface{} instead
+// of specialized structures.  It is needed becuase JSON library is
+// case insensitive when matching json values to struct fields
+type ciConfig struct {
+	conf   map[string]interface{}
+	fromLC map[string]string
+}
+
+func newCiConfig(conf map[string]interface{}) ciConfig {
+	return ciConfig{
+		conf:   conf,
+		fromLC: lcMap(conf),
+	}
+}
+
+func lcMap(conf map[string]interface{}) map[string]string {
+	fromLC := make(map[string]string)
+	for key, _ := range conf {
+		fromLC[strings.ToLower(key)] = key
+	}
+	return fromLC
+}
+
+// get gets a key, returns nil if the key doesn't exist
+func (c ciConfig) get(origkey string) interface{} {
+	lckey := strings.ToLower(origkey)
+	key, ok := c.fromLC[lckey]
+	if !ok {
+		return nil
+	}
+	return c.conf[key] // should not panic, key should exist
+}
+
+func (c ciConfig) set(origkey string, val interface{}) {
+	lckey := strings.ToLower(origkey)
+	key, ok := c.fromLC[lckey]
+	if ok {
+		c.conf[key] = val
+	}
+	c.fromLC[lckey] = origkey
+	c.conf[origkey] = val
+}
+
+func (c ciConfig) del(origkey string) {
+	lckey := strings.ToLower(origkey)
+	key, ok := c.fromLC[lckey]
+	if ok {
+		delete(c.conf, key)
+	}
+}
--- a/ipfs-5-to-6/migration/config_conv.go
+++ b/ipfs-5-to-6/migration/config_conv.go
@@ -1,0 +1,158 @@
+package mg5
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"reflect"
+)
+
+// convFunc does an inplace conversion of the "datastore"
+// configuration (represted as a map[string]interface{}) from one
+// version to another
+type convFunc func(ds ciConfig) error
+
+// convertFile converts a config file from one version to another, the
+// converted config is stored in
+func convertFile(orig string, new string, convFunc convFunc) (ciConfig, error) {
+	in, err := os.Open(orig)
+	if err != nil {
+		return ciConfig{}, err
+	}
+	out, err := os.Create(new)
+	if err != nil {
+		return ciConfig{}, err
+	}
+	return convert(in, out, convFunc)
+}
+
+// convert converts the config from one version to another, returns
+// the converted config as a map[string]interface{}
+func convert(in io.Reader, out io.Writer, convFunc convFunc) (ciConfig, error) {
+	data, err := ioutil.ReadAll(in)
+	if err != nil {
+		return ciConfig{}, err
+	}
+	confMap := make(map[string]interface{})
+	err = json.Unmarshal(data, &confMap)
+	if err != nil {
+		return ciConfig{}, err
+	}
+	conf := newCiConfig(confMap)
+	ds, _ := conf.get("datastore").(map[string]interface{})
+	if ds == nil {
+		return ciConfig{}, fmt.Errorf("Datastore field missing or of the wrong type")
+	}
+	err = convFunc(newCiConfig(ds))
+	if err != nil {
+		return ciConfig{}, err
+	}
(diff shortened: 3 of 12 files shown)


kevina:
@whyrusleeping right now the conversion code will leave the backup file `config-v5` (or `config-v6` when downgrading) in case there is a problem.

ipfs-5-to-6/migration/config_conv.go
```diff
@@ -0,0 +1,154 @@
...
+		return err
+	}
+	out, err := os.Create(new)
+	if err != nil {
+		return err
+	}
+	return convert(in, out, convFunc)
+}
+
+func convert(in io.Reader, out io.Writer, convFunc func(ds map[string]interface{}) error) error {
```
whyrusleeping: I would separate the code that does json marshaling from the code thats actually doing the conversion. Maybe have the convert method take in a map[string]interface, and return one as well, then the function that calls it handles all the json marshaling fun

ipfs-5-to-6/migration/config_conv.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func ver6to5(ds map[string]interface{}) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("incompatible config detected, downgrade not possible: %v", r.(error))
+		}
+	}()
+	fromLC := lcMap(ds)
+	spec := ds[fromLC["spec"]].(map[string]interface{})
```
whyrusleeping: i would do type asserts here, if the users config is messed up this may panic.

whyrusleeping: ( i know we're catching it, but its better to have a slightly more descriptive error message than the panic )

kevina: @whyrusleeping I am caching the panic to avoid having to write error checking code for every bit of the conversion. Do you just want a better error message if the "spec" field is missing or the wrong type?

ipfs-5-to-6/migration/config_conv.go
kevina: @whyrusleeping I am confused, that is basically what I am doing.  Convert calls convFunc that does an inplace conversion on `map[string]interface`.  Maybe some better documentation to make it clear what is going on.

whyrusleeping: @kevina ah, yeah, maybe some better documentation then

ipfs-5-to-6/migration/config_conv.go
whyrusleeping: @kevina yeah, checking that would be good. fs-repo-migrations should try to be as safe as possible, and also be very clear on what went wrong when something goes wrong.

whyrusleeping:
@kevina whats the status here? This is feature complete except for the 'repo id' thing, right?

kevina:
It is feature complete.  It needs better testing.  There is no need for the conversion code to write the disk id/spec file out as that will be created when the database is first opened if it doesn't exist. See https://github.com/ipfs/go-ipfs/pull/3575#issuecomment-313568157.

kevina:
I need to rebase to clean up my commits but I consider this done now.

Note, that I wrote some unit tests that don't seam to be run by the build server as it doesn't seam to run `go test`.

whyrusleeping:
will need to bump the go version in CI to 1.8

kevina:
@whyrusleeping okay, I'll do that, what about running `go test`.

ipfs-5-to-6/migration/config_conv.go
```diff
@@ -0,0 +1,155 @@
...
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("incompatible config detected, downgrade not possible: %v", r.(error))
+		}
+	}()
+	spec := ds.get("Spec").(map[string]interface{})
+	if spec == nil {
+		return fmt.Errorf("Datastore.Spec field missing or not a json object, can't downgrade")
+	}
+	mounts := spec["mounts"].([]interface{})
```
whyrusleeping: the recover is a good fallback safeguard, but i would still like to be able to catch these specific issues. If we check this assertion, we can tell the user that their mounts settings is not an array.

ipfs-5-to-6/migration/config_conv.go
```diff
@@ -0,0 +1,155 @@
...
+	sync := true
+	for _, mount := range mounts {
+		switch mountpoint := mount.(map[string]interface{})["mountpoint"].(string); mountpoint {
+		case "/blocks":
+			sync = mount.(map[string]interface{})["child"].(map[string]interface{})["sync"].(bool)
+			blocks = mount
+		case "/":
+			root = mount
+		default:
+			return fmt.Errorf("unknown mountpoint")
```
whyrusleeping: I would be more specific about which mountpoint is unknown:
```
return fmt.Errorf("unknown mountpoint: %q", mountpoint)
```

ipfs-5-to-6/migration/config_conv.go
kevina: If that was the case I don't think it will possible to represent that configuration with the older config style and hence we won't be able to downgrade.

kevina: I can add this specific check.

The reason for the fallback is if I check each and every possible error the code will be 3 times as large as there is a lot that can go wrong here.

kevina:
@whyrusleeping this needs a rebase but after that should be good to go.  Seemingly unrelated tests are failing on the build server.

ipfs-5-to-6/migration/ci_config.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"strings"
+)
+
+// ciConfig is a case insensitive map to represent a IPFS
+// configuration when unmarshaled as a map[string]interface{} instead
+// of specialized structures.  It is needed becuase JSON library is
+// case insensitive when matching json values to struct fields
+	type ciConfig struct {
```
Stebalien: `go fmt`

.travis.yml
```diff
@@ -8,14 +8,14 @@ os:
...
+  - 1.8
 
 env:
   - TEST_VERBOSE=1
 
 script:
-  - cd sharness
-  - make
+  - make -C sharness
+  - go test ./ipfs-5-to-6/migration/
```
Stebalien: Why are we testing this migration separately here?

kevina: I commented on that above:

> Note, that I wrote some unit tests that don't seam to be run by the build server as it doesn't seam to run go test.

Stebalien: Maybe add a `test` target to the Makefile and invoke that?

```make
test:
	$(MAKE) -C sharness
	go test ./...
```

Stebalien: Really,  we should add a call to `go vet ./...` but that fails on the current repo so we can do that in a different PR.

kevina: I can add a test target, but "go test ./..." fails locally for me, which is why I only added my tests.

Stebalien:
Oh... Can you just add a todo and file an issue?

whyrusleeping:
@kevina I'm back at it now, Lets get this moving again :) I'm feeling like we're very close

whyrusleeping:
@lgierth also wants some stuff added to this migration for the configs. I think we should get this PR merged, and then we can do a separate PR for those changes.

whyrusleeping:
Having problems reverting:
```
why@manwe ~/g/s/g/i/fs-repo-migrations> ipfs-5-to-6 -revert -path="$HOME/.ipfs"
reverting migration
ver6to5 done!
why@manwe ~/g/s/g/i/fs-repo-migrations> ipfs daemon --offline
Initializing daemon...
Adjusting current ulimit to 2048...
Successfully raised file descriptor limit to 2048.
Error: Your programs version (5) is lower than your repos (6).
Please update ipfs to a version that supports the existing repo, or run
a migration in reverse.

See https://github.com/ipfs/fs-repo-migrations/blob/master/run.md for details.
```

kevina:
@whyrusleeping can you check if the config looks good after the downgrade, if you manually fix the version string are things okay?

whyrusleeping:
@kevina yeah, the config looks good to me, and manually changing the repo version seems to fix the issue.

kevina:
@whyrusleeping fixed

kevina:
`t0091-4-to-5-error-recovery.sh` is failing and I'm not sure why, I don't see how my code could of broken this but you never know...

kevina:
@whyrusleeping this should be good to go now.

A few things to note (not necessary a problem)

1) The sharness test are very basic, it simply upgrades and than reverts.  The go unit tests, however, make sure the conversion is being done correctly.
2) The conversion code will leave the backup file `config-v5` (or `config-v6` when downgrading) in place.  I figured it would be good to keep in case there was problem
3) `t0091-4-to-5-error-recovery.sh` is failing, but I don't see how my code could be causing this

kevina:
@whyrusleeping https://github.com/ipfs/fs-repo-migrations/pull/69 should probably go in first to fix `t0091-4-to-5-error-recovery.sh`.

whyrusleeping:
@kevina merged #69 wanna rebase this one? Other than the test failures (which should be resolved after rebase) this should be good to go, at least for testing and getting people to try out the new features

kevina:
@whyrusleeping One step ahead of you.  Just rebased. :)

whyrusleeping:
and look how green everything is!

whyrusleeping:
I'm gonna try it out a bit more locally, and then merge :)

kevina:
@whyrusleeping thanks

whyrusleeping:
alright, tested this out a bit, it looks good to me. Lets merge it so people can try out the datastore config PR more easily.

PR #76, opened 2018-06-16
Moved gx deps to top level.

kevina:


Stebalien:
@whyrusleeping you had opinions on this?

hsanjuan:
we dont use gx anymore

PR #80, opened 2018-09-09
[DRAFT] named pins migration

Voker57:
This is migration for  https://github.com/ipfs/go-ipfs/pull/4757, untested cause I haven't been able to figure out how to compile migrations or run then without published go-ipfs versions.


hsanjuan:
I don't know if a PR can be set to "draft" mode, but given this is old and the effort frozen,  I am going to close this and it can always be re-opened if need arises.

PR #89, opened 2019-12-17
Update bootstrappers

dirkmc:
As part of [Fixing Bootstrapping](https://github.com/ipfs/go-ipfs/issues/6797) we need to add a repo migration that
- removes bootstrap nodes with small keys
- adds new nodes with appropriately large keys

TODO:
- [ ] write sharness tests


ipfs-7-to-8/migration/config_conv_test.go
```diff
@@ -0,0 +1,120 @@
...
+	]
+}
+`
+
+var expectedMigrateForward = []string{
+	"/dnsaddr/bootstrap.libp2p.io/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN",
+	"/dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa",
+	"/dnsaddr/bootstrap.libp2p.io/p2p/QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb",
+	"/dnsaddr/bootstrap.libp2p.io/p2p/QmcZf59bWwK5XFi76CZX8cbJ4BhTzzA3gU1ZjYZcYW3dwt",
+	"/ip4/104.131.131.82/tcp/4001/p2p/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ",
```
ribasushi: @dirkmc is it intentional that you are looking only for the `.82` host here, but above ( line 96 ) there are 2 "old style" hosts?

Approving, as the rest looks good and the above is likely intentional

dirkmc: Good question - yes, in the migrate forward we move from `oldConfig` to `expectedMigrateForward`. In the migrate backwards we move from `newConfig` to `expectedMigrateBackward` for which I wanted a test that includes one peer that is "known" (in the list of peers with acceptably large keys) and one that is not in that list, to make sure both get through.

aschmahmann:
@dirkmc overall seems reasonable a couple comments/thoughts.

1) Repo migrations look really painful, it also looks like there's a bunch of boilerplate required to do a really simple change
2) Doing a repo migration to update bootstrappers seems a little wacky. I understand that if we need this now then we've just go to do it, but is there a better way to deal with this? If we have to do a migration anyway maybe we should add a `UseDefaults` boolean and just remove all of our bootstrappers, this will let us push updates in code instead of forcing repo migrations.

ipfs-7-to-8/migration/config_conv.go
```diff
@@ -0,0 +1,177 @@
...
+func ver7to8(bootstrap []string) []string {
+	// Make sure the dnsaddrs bootstrap peers are included
+	var res []string
+	for _, p := range dnsBootstrapPeers {
+		res = append(res, fmt.Sprintf("/dnsaddr/%s/p2p/%s", dnsAddr, p))
+	}
+
+	// Filter out peers that we added above, or that have an ID known to belong
+	// to a peer with a small key
+	for _, addr := range bootstrap {
```
Stebalien: We should do the conversion if, and only if:

1. There are no dnsaddr keys in the list.
2. There is at least one small key in the list.

Basically, we _don't_ want to override the user's choice. We want to _replace_ the small keys with the new bootstrappers (if present), not add the new bootstrappers regardless.

Stebalien: For example, the user may be on a private network.

ipfs-7-to-8/migration/config_conv.go
```diff
@@ -0,0 +1,177 @@
...
+	fixed, err := json.MarshalIndent(confMap, "", "  ")
+	if err != nil {
+		return err
+	}
+	out.Write(fixed)
+	out.Write([]byte("\n"))
+	return nil
+}
+
+func ver7to8(bootstrap []string) []string {
```
Stebalien: If possible, let's try to avoid reordering. _Especially_ if we don't end up modifying the set of bootstrappers.

ipfs-7-to-8/migration/config_conv.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+		oldPeerIDs[pid] = struct{}{}
+	}
+
+	// Filter out old addresses added above, and addresses with the new DNS
+	// addresses
+	for _, btAddr := range bootstrap {
+		pid, err := getAddrPeerID(btAddr)
+		if err == nil {
+			if _, ok := oldPeerIDs[pid]; !ok && !strings.Contains(btAddr, dnsAddr) {
```
Stebalien: Let's not _remove_ the new bootstrappers when we revert. We just want to re-add the old ones if, and only if, our dnsaddr bootstrappers have been specified.

ipfs-7-to-8/migration/config_conv.go
```diff
@@ -0,0 +1,177 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/ipfs/go-cid"
```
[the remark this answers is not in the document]
dirkmc: I used a tool to generate the vendor directory, but it ends up being very big because of all the crypto dependencies. I only use go-cid to verify the peer ID, so alternatively I could copy some of the go-cid code that does verification. @Stebalien which do you think is the better approach?

ipfs-7-to-8/migration/config_conv.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		return err
+	}
+	confMap := make(map[string]interface{})
+	if err = json.Unmarshal(data, &confMap); err != nil {
+		return err
+	}
+	bootstrapi, _ := confMap["Bootstrap"].([]interface{})
+	if bootstrapi == nil {
+		bootstrapi, _ := confMap["bootstrap"].([]interface{})
```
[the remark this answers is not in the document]
dirkmc: A lot of this code is copied from https://github.com/ipfs/fs-repo-migrations/tree/master/ipfs-5-to-6/migration
There's a comment there implying that keys are not case sensitive:
https://github.com/ipfs/fs-repo-migrations/blob/46ecb9d16dbc5bcbc420369bbe247ac5f13a2268/ipfs-5-to-6/migration/ci_config.go#L7-L14

dirkmc:
@aschmahmann you're right that there's a lot of boilerplate. Much of this code is copied from https://github.com/ipfs/fs-repo-migrations/tree/master/ipfs-5-to-6 (and much of that is copied from earlier version changes)
Because the code is intended to be used at one point in time and remain static thereafter I'm not sure if this is a big issue.
With respect to doing a repo migration to update bootstrappers I agree it is a little strange. It may be worth talking through a better way to apply config-only changes if this is something that happens often.

aschmahmann:
@dirkmc as I mentioned in the issue I linked we really need to up our config file game so that we can deal with default values. Currently we just give people a config file on init and the only way to upgrade it is a repo migration or having users manually change the files.

We have some experimental features that can be enabled optionally via config (e.g. pubsub) that when we want to enable by default (we already do) we don't have any choices other than modifying the user's config file (e.g. via a repo migration)

ipfs-7-to-8/migration/config_conv.go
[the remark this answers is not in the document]
Stebalien: Hm. Actually, looking at that code, we shouldn't be using CID. _Technically_, peer IDs can be decoded as CIDs. However, they're quite the same thing.

I'd actually just remove that code. We should just check "does the address match `/(p2p|ipfs)/QmTargetPeerID$`?". Then we can skip all deps.

dirkmc:
@Stebalien I think I've addressed all your suggestions, let me know if the logic looks right to you now

ipfs-7-to-8/migration/config_conv_test.go
dirkmc: I updated the var names to make it a little clearer what's going on, and added some more explicit tests

ipfs-7-to-8/migration/config_conv.go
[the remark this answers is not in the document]
Stebalien: (╯°□°)╯︵ ┻━┻

ipfs-7-to-8/migration/config_conv.go
```diff
@@ -0,0 +1,214 @@
...
+
+	// Filter out dnsaddr peers that we added above, or that have an ID known
+	// to belong to a peer with a small key.
+	// If we don't recognize the address format (err != nil), assume that the
+	// user changed it for a reason and leave it in there.
+	for _, addr := range bootstrap {
+		if isDNSAddr, err := isDNSBootstrapPeer(addr); !isDNSAddr || err != nil {
+			if isSmall, err := isSmallKeyPeer(addr); !isSmall || err != nil {
+				// Replace /ipfs with /p2p
+				addr = strings.Replace(addr, "/ipfs", "/p2p", -1)
```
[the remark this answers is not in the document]
dirkmc: Ah yeah that's true - I made a fix, I just want to confirm that's how you were imagining it, does that look right now?

Stebalien:
@dirkmc how's your internet? To test this, I think we need to:

Build https://github.com/ipfs/go-ipfs/pull/6814 with https://github.com/ipfs/distributions:
1. `ipfs get -o releases /ipns/dist.ipfs.io`. This will download 10GiB of crap and add it to your local IPFS repo, FYI. However, you'll only have to download it once.
1. (`./dist.sh add-version go-ipfs $HASH_OF_THAT_PR`).
2. `make publish` (let's call the final hash `MY_DIST_PATH`

(checkout the readme in github.com/ipfs/distributions, it should walk you through this process)

This will publish a local release of IPFS with `$HASH_OF_THAT_PR` as the version. Note: Given that we normally use, you know, sane version numbers, I'm not sure if this is going to actually work. We may need a step 1.5 that moves the built release to some real version (v0.5.0-beta?). If so, you'll probably need to rename a bunch of files and edit the dist.json file.

Once you have everything built, you can write the sharness test. Model it after `t0110-migration-6-to-7.sh`:
  1. Set `IPFS_DIST_PATH` to `MY_DIST_PATH`
  2. Use `$HASH_OF_THAT_PR` as the new version and v0.4.22 as the old version.

If everything works, pin `MY_DIST_PATH` using pinbot. We don't need to _publish_ this release, we just need it to stick around.

Stebalien:
Merging as there's no reason to leave this sitting in a PR. It won't be used until we bump the repo version and cut a release.

dirkmc:
Thanks 👍 

I messed around with trying to get sharness tests working with a temp release but was having trouble with the old, vendored dependencies (eg ipfs-update).

It may make sense to wait till we have a RC so as to avoid having to cut a temp release and the versioning issues that raises when trying to write sharness tests.

Stebalien:
Yeah, there's no point in wasting time on this. I'm going to see if I can clean up the deps at all but I won't spend too much time at it.

PR #94, opened 2020-01-31
Gxed Badger hack: use the same net/trace everywhere

hsanjuan:
net/trace cannot be imported on different versions from different modules.

Additionally, we also cannot export ExpVars with the same names from two different
packages (one being gxed-badger and the other being the go-modded one). Since those
two packages have different import paths, go considers them different and runs
init() twice, causing the issues.


hsanjuan:
@Stebalien I have isolated this change here so that the migrations PR is cleaner.

go.mod
```diff
@@ -1,3 +1,5 @@
 module github.com/ipfs/fs-repo-migrations
 
 go 1.13
+
+require golang.org/x/net v0.0.0-20200114155413-6afb5195e5aa
```
Stebalien: Let's still vendor this, just share it. The migrations should build even if all external deps go down.

Stebalien:
If we use `go mod vendor`, we need to make sure to build with `go build -mod=vendor`. Although I guess this + go.sum meets the "works if everything else goes down" and "doesn't" change requirements.

Stebalien:
That's why I avoided using `go mod vendor` for all of our other dependencies (e.g., gx, etc.). `go build` will still fail if the source goes down unless you specify `go build -mod=vendor`.

go.mod
hsanjuan: ok, I vendored this, and I made sure .travis.yml is building with vendored deps. If that works, it means we should be able to build migrations offline if needed.

hsanjuan:
@Stebalien you commented while I was vendoring all the things properly. See now?

hsanjuan:
So now travis will catch if things not vendored

PR #95, opened 2020-01-31
 Migration 8-9: migrate CIDsv1 to raw multihashes

hsanjuan:
This adds the migration to convert CIDsV1 in the datastore "/blocks" namespace to raw multihashes.

Testing is missing, but initial testing on my local repo (~5000 of such CIDs) seems to work on both flatfs (slow) and badger (really fast).

To be rebased on #94 . More comments inlined.


ipfs-8-to-9/migration/go.mod
```diff
@@ -0,0 +1,21 @@
+module github.com/ipfs/fs-repo-migrations/ipfs-8-to-9/migration
```
hsanjuan: I am not sure there is a good way of having self-contained vendored-like dependencies  after the issues #94 . Likely net/trace needs to be kept aligned for all migrations in all submodules. Annoying but probably do-able anyways.

ipfs-8-to-9/migration/go.mod
```diff
@@ -0,0 +1,21 @@
...
+
+go 1.13
+
+require (
+	github.com/hsanjuan/go-libp2p-http v0.0.2 // indirect
```
hsanjuan: This should be libp2p/go-libp2p-http. Need to depend on ipfs-master-fake-tagged version.

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,110 @@
...
+// Versions returns the current version string for this migration.
+func (m Migration) Versions() string {
+	return "8-to-9"
+}
+
+// Reversible returns false. This migration cannot be reverted, as we do not
+// know which raw hashes were actually CIDv1s. However, things should work all
+// the same as they will be treated as CIDv0s in old versions anyways.
+func (m Migration) Reversible() bool {
+	return false
```
hsanjuan: Options:

* A: Say this migration is not reversible.
* B: Say this migration is reversible but treat it as no-op because things pretty much work as before if the user was to run with the previous go-ipfs version.
* C: Write down the list of all CIDs that have been migrated and re-use that to revert the migration.

I vote B unless I'm missing something.

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,110 @@
...
+	dsc, err := fsrepo.AnyDatastoreConfig(cfg.Datastore.Spec)
+	if err != nil {
+		return err
+	}
+
+	dstore, err := dsc.Create(opts.Path)
+	if err != nil {
+		return err
+	}
+	defer dstore.Close()
```
hsanjuan: Opted to go this way as I am not allowed to use Open() because ipfs repo version is lower than expected. If we depend on a newer version, might use `Open()` directly. But I'm thinking I'll still need to load the plugins.

Unfortunately did not find a pretty way to parse configuration manually and open datastores by hand. The fact that specs are `map[string]interface{}` with very different possibilities inside does not help.

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// TODO: assuming the user has not modified this
+	blocks := namespace.Wrap(dstore, ds.NewKey("/blocks"))
```
hsanjuan: Could actually process the root namespace and migrate everything that ends with a CIDv1. Do we want that? This would, afaik, migrate providers too.

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,110 @@
...
+	log.Log("updated version file")
+
+	return nil
+}
+
+// Revert attempts to undo the migration.
+func (m Migration) Revert(opts migrate.Options) error {
+	// TODO: Consider a no-op revert though
+	return errors.New("This migration cannot be reverted")
+}
```
hsanjuan: See comment above.

ipfs-8-to-9/migration/swapper.go
```diff
@@ -0,0 +1,165 @@
...
+
+// NWorkers sets the number of batching threads to run
+var NWorkers int = 4
+
+// CidSwapper reads all the keys in a datastore and replaces
+// them with their raw multihash.
+type CidSwapper struct {
+	Store ds.Batching // the datastore to migrate.
+}
+
```
hsanjuan: I want to eventually use this in cluster. Thus, exported.

ipfs-8-to-9/migration/swapper.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/ipfs/go-datastore"
+	ds "github.com/ipfs/go-datastore"
+	"github.com/ipfs/go-datastore/query"
+	dshelp "github.com/ipfs/go-ipfs-ds-help"
+	"github.com/multiformats/go-multihash"
+)
+
+// SyncSize specifies how much we batch data before committing and syncing
+// Increasing this number may result
+var SyncSize uint64 = 10 * 1024 * 1024 // 1MiB
```
hsanjuan: TODO Fix: actually 10MiB.

ipfs-8-to-9/migration/swapper.go
```diff
@@ -0,0 +1,165 @@
...
+		if err != nil {
+			log.Error("swapping %s for %s: %s", oldKey, newKey, err)
+			errored++
+			continue
+		}
+		swapped++
+		curSyncSize += size
+
+		// Commit and Sync if we reached SyncSize
+		if curSyncSize >= SyncSize {
```
hsanjuan: Alternatively we could batch based on number of keys migrated. Total size seemed a better option though.

main.go
```diff
@@ -29,6 +31,8 @@ var migrations = []gomigrate.Migration{
 	&mg4.Migration{},
 	&mg5.Migration{},
 	&mg6.Migration{},
+	&mg7.Migration{}, // update bootstrappers
+	&mg8.Migration{}, // converts CIDv1s to raw multihashes
```
hsanjuan: This enables the bootstrappers migration.

Stebalien:
(you checked in a binary)

ipfs-8-to-9/migration/migration.go
Stebalien: Well, I assume we'd depend on the _current_ version of go-ipfs (old repo).  Also note, we can just set `fsrepo.RepoVersion`. Take a look at `Revert` in 6-to-7.

However, what you're doing here definitely works and is probably better (no measure datastore).

ipfs-8-to-9/migration/migration.go
Stebalien: Hm. I'd wouldn't. We'll delete old providers separately. I don't want to migrate _everything_ that looks like a CID.

Stebalien: Real question: do we need to migrate the filestore? Probably.

ipfs-8-to-9/migration/swapper.go
Stebalien: I agree, total size seems like the way to go.

Stebalien: Oh, wait. I'm not entirely sure if this is correct. The idea behind batching was to:

1. Add N new blocks in a batch.
2. Commit and Sync.
4. Remove the old versions in a batch
4. Commit and Sync.

I don't think the datastores (e.g., badger) guarantee ordering of events until a sync happens. Otherwise, we'd just sync once at the end and be done with it.

TL;DR: periodic syncing doesn't really buy us anything.

ipfs-8-to-9/migration/swapper.go
```diff
@@ -0,0 +1,165 @@
...
+
+// swap swaps the old for the new key in a single transaction.
+func (cswap *CidSwapper) swap(old, new ds.Key) (uint64, error) {
+	// Unfortunately grouping multiple swaps in larger batches usually
+	// results in "too many open files" errors in flatfs (many really
+	// small files) . So we do small batches instead and Sync at larger
+	// intervals.
+	// Note flatfs will not clear up the batch after committing, so
+	// the object cannot be-reused.
+	batcher, err := cswap.Store.Batch()
```
Stebalien: Hm. We could try to detect flatfs. Not sure if there's an easy way to do this now. We might need to extend the datastore interface.

ipfs-8-to-9/migration/go.mod
hsanjuan: > I'd manually vendor

That seems even more work.

> That, or move away from the single binary approach

I think this is the sanest option. The general binary should call the other binaries rather than importing. Also, I can imagine ipfs only downloading the migrations it needs to download rather than a big file with all.

However, we can do that outside of this PR things are not broken right now.

ipfs-8-to-9/migration/migration.go
hsanjuan: Yeah, C it is.

ipfs-8-to-9/migration/migration.go
hsanjuan: > no measure datastore

Sure? Isn't the Spec wrapping in measure ds anyways (for flatfs at least)?

ipfs-8-to-9/migration/migration.go
hsanjuan: Yes, we need, I guess there is a namespace that I haven't found out yet.

That said, we do want to migrate providers too, but can scope for their namespace.

ipfs-8-to-9/migration/swapper.go
hsanjuan: > I don't think the datastores (e.g., badger) guarantee ordering of events until a sync happens

You mean you can be left in a state where the last write operation in a transaction succeeded but the previous ones were lost if something happens mid-sync? Well that's shitty..

The problem with `adding N blocks` to a batch and commit is that, if blocks are very small, we will be batching really small sizes. And we need to keep N low enough to not trigger `too many files opened` in flatfs, so it is going to be a low N given there are 4 workers.

How about we do two full passes, one adding all the new blocks, syncing by size as now, then another removing the old blocks?

ipfs-8-to-9/migration/go.mod
Stebalien: > I think this is the sanest option. The general binary should call the other binaries rather than importing. Also, I can imagine ipfs only downloading the migrations it needs to download rather than a big file with all.

I agree. Even better, we can move the main "which migrations do I run" logic into go-ipfs itself.

> However, we can do that outside of this PR things are not broken right now.

That will back us into a corner where we _won't_ be able to add a new migration till we fix this. I'm not willing to take a risk like that unless we're up against a wall.

ipfs-8-to-9/migration/migration.go
Stebalien: Ah, interesting. We're measuring multiple times. One for "global" datastore stats, one per datastore type. We should reconsider that (https://github.com/ipfs/go-ipfs/issues/6877).

(not all that important for the current change)

ipfs-8-to-9/migration/migration.go
Stebalien: > That said, we do want to migrate providers too, but can scope for their namespace.

Why? They'll expire in a day.

ipfs-8-to-9/migration/go.mod
hsanjuan: Ok, many moving pieces. I propose:

1) Finish things here as they are, leave go-ipfs with a working tested migration-set.
2) Work on splitting migration binaries and teaching go-ipfs how to use them next. We will need to publish migration binaries separately (migrations.ipfs.io?)/

We need to avoid snowballs that end up blocking releases or leaving a lot never-merged code behind. The fact that there are 3 pending migrations is a large snowball already.

ipfs-8-to-9/migration/migration.go
hsanjuan: Because it is effortless to migrate another namespace, but if you tell me nothing is going to break in those 24h, I can live with that.

ipfs-8-to-9/migration/go.mod
Stebalien: Ok... Then let's vendor these deps for now.

ipfs-8-to-9/migration/migration.go
Stebalien: If it's easy, we can go ahead and do it anyways. My main concern is that I don't want to sit their migrating those records

If we're migrating them anyways, I'd consider (don't spend too much time at this) checking to see if they're expired and delete them up-front (should be faster than re-writing them).

ipfs-8-to-9/migration/swapper.go
Stebalien: > You mean you can be left in a state where the last write operation in a transaction succeeded but the previous ones were lost if something happens mid-sync? Well that's shitty..

Not in a _transaction_. In a transaction, we'll be fine. However, batches aren't _technically_ transactions. They are in badger (implementation detail), but not in flatfs.

On the other hand, this logic is _probably_, technically safe in most/all existing case.

* Badger: Batches are actually transactions.
* Flatfs: When applying a batch, we put first, sync, then delete.
* Levelds: I think the batch is applied in-order then synced at the end.

> How about we do two full passes, one adding all the new blocks, syncing by size as now, then another removing the old blocks?

That would be safe. However, my concern there is that we'll double the size of the datastore (worst case).

Note: we don't actually need to use _batches_. We could also:

1. Add up to N megabytes.
2. Sync.
3. Remove the keys we just added.
4. Repeat.

Basically, your solution but we do it iteratively (without batches).

hsanjuan: So I added up to SyncSize, then Sync, then delete, then Sync, then repeat.

ipfs-8-to-9/migration/swapper.go
hsanjuan: If the user has sync enabled, then it gets a nice sync migration. If it is disabled, then both flatfs and badger should be faster. I am not sure batching gives us much on async scenarios but I might be wrong?

ipfs-8-to-9/migration/go.mod
hsanjuan: I will have to add the `vendor` folder at the end or in a different PR, since it pollutes this review.

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	if err := plugins.Initialize(); err != nil {
+		return nil, fmt.Errorf("error initializing plugins: %s", err)
+	}
+
+	if err := plugins.Inject(); err != nil {
+		return nil, fmt.Errorf("error injecting plugins: %s", err)
+	}
+
```
hsanjuan: TODO: potentially `fsrepo.Open` when linked to a newer ipfs tag.

ipfs-8-to-9/migration/go.mod
```diff
@@ -0,0 +1,19 @@
+module github.com/ipfs/fs-repo-migrations/ipfs-8-to-9/migration
+
+go 1.13
+
+require (
+	github.com/ipfs/fs-repo-migrations v1.4.0
+	github.com/ipfs/go-cid v0.0.5
+	github.com/ipfs/go-datastore v0.3.1
+	github.com/ipfs/go-filestore v0.0.4-0.20200203163551-8cc45055f624 // indirect
+	github.com/ipfs/go-ipfs v0.4.22-0.20200130064341-6750ee973e2a
```
hsanjuan: TODO: link to newer after https://github.com/ipfs/go-ipfs/pull/6816 has the right dep-set and is green.

ipfs-8-to-9/migration/migration.go
hsanjuan: @Stebalien I did not realize, the problem is there is actually potentially a very large number of provider blocks to migrate? Might also treat this as an additional migration later. Let me look up some code, there are some things I'm unfamiliar with here. Let's keep this migration as only doing `blocks`.

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,242 @@
...
+		log.Error("failed to write version file")
+		return err
+	}
+
+	log.Log("reverted version file to version 8")
+	f.Close()
+	err = os.Remove(backupPath)
+	if err != nil {
+		log.Error("could not remove the backup file, but migration worked: %s", err)
+	}
```
hsanjuan: TODO: Not sure if migration should end up in error if file cannot be removed. Not sure if file should be removed after revert.

ipfs-8-to-9/migration/swapper.go
hsanjuan: The delete part is effectively a flatfs Batch, which makes me a bit scared about the file descriptor limit, but I haven't hit issues.

ipfs-8-to-9/migration/migration.go
Stebalien: It might be, I don't actually know.

ipfs-8-to-9/migration/swapper.go
Stebalien: We don't actually support async updates in flatfs. Well, you _can_ disable "sync", but `Sync` won't do anything. To implement it correctly, we'd need to keep some list of unsynced files.

> I am not sure batching gives us much on async scenarios but I might be wrong?

Batching will likely help with flatfs regardless because it does all writes (open, write, close, etc.) in parallel. But we may be over-optimizing.

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,242 @@
...
+	defer buf.Flush()
+
+	swapCh := make(chan Swap, 1000)
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for sw := range swapCh {
+			fmt.Fprint(buf, sw.Old.String()+","+sw.New.String()+"\n")
```
Stebalien: To be safe, I think we need to write this _first_. We might want to:

1. Walk the datastore, writing down every CIDv1 we come across.
2. Do it again, doing the swaps.

Stebalien: I.e., write ahead logging: write what we're going to do (what we're going to migrate), do it (swap), then say we did it (update the repo version).

ipfs-8-to-9/migration/migration.go
Stebalien: I agree. If we fail to remove it, leave it. However, if we're trying to go _forward_, we need to make sure to replace/remove this file before we even start.

ipfs-8-to-9/migration/migration.go
Stebalien: Note: we also don't need to record old/new. We can derive the multihash from the CID (and cut the filesize in half).

ipfs-8-to-9/migration/swapper.go
```diff
@@ -0,0 +1,265 @@
...
+	// Delete all the old keys
+	for _, o := range sw.toDelete {
+		if err := sw.store.Delete(o); err != nil {
+			return err
+		}
+	}
+	sw.toDelete = nil
+
+	// Sync again.
+	err = sw.store.Sync(ds.NewKey("/"))
```
Stebalien: Thinking about this a bit, we actually don't need this second sync. We just need adds to persist before we delete, a final sync at the very end will sync the any unfinished deletes.

ipfs-8-to-9/migration/swapper.go
Stebalien: Should be fine. Delete doesn't open a file descriptor (just sends the path itself down to the kernel).

ipfs-8-to-9/migration/swapper.go
hsanjuan: Let's leave it "unoptimized" for now. I will add some verbosity so that the user sees progress. We can come back to this if it's really really bad.

ipfs-8-to-9/migration/migration.go
hsanjuan: Current version renames it, and returns error if that does not work, but the migration should be ok, and if it is moving forward again, and the user did not care to touch the file, it will just append to it, which is still fine even if it was used for another revert (will just take longer).

I really hope no one runs a migration followed by a revert which actually only fails on the very last step of renaming a file, anyways.

ipfs-8-to-9/migration/migration.go
hsanjuan: > To be safe, I think we need to write this _first_. We might want to:
> 
>     1. Walk the datastore, writing down every CIDv1 we come across.
> 
>     2. Do it again, doing the swaps.

Ok, done.

> Note: we also don't need to record old/new. We can derive the multihash from the CID (and cut the filesize in half).

Ok, done this too.

ipfs-8-to-9/migration/swapper.go
hsanjuan: Yeah, I did it here so that I don't have to do a final sync in two places, but I'll change that.

Stebalien:
So, I have some bad news, reverts going to get more complicated.

If we already have a block as CIDv1 and CIDv0, the revert will delete the CIDv0. We can easily fix this by either not deleting on revert or by checking to see if we already have the block before marking it as something we've swapped.

Unfortunately, the second issue is much worse: If we upgrade, add some files with CIDv1, then downgrade, we'll GC these files. It'll still be possible to read the files given that we check both V0 and V1 CIDs on read. The only solution I can think of is to revert by walking the pinset/mfs. If we do that, with https://github.com/ipfs/go-ipfs/blob/master/thirdparty/cidv0v1/blockstore.go, we'll avoid GCing anything we want to keep.

Thoughts? I'd rather not say "thou shalt not revert" given all the other features and potential issues with this release, but this is getting to be a _very_ complicated process.

hsanjuan:
> If we already have a block as CIDv1 and CIDv0, the revert will delete the CIDv0. We can easily fix this by either not deleting on revert or by checking to see if we already have the block before marking it as something we've swapped.

ok, doable somehow. We can tag certain migrations as "keep" and don't delete the blocks during revert.

> Unfortunately, the second issue is much worse...

This is getting hell-ish. The GC code (I assume) has the necessary code to list all the CIDs under pinset and mfs, so maybe it is not overly complicated to add the revert of those blocks as an extra step for the revert process?

Stebalien:
> ok, doable somehow. We can tag certain migrations as "keep" and don't delete the blocks during revert.

SGTM.

> This is getting hell-ish.

I agree. I'm starting to think we should put this in its own release instead of squeezing it into 0.5.0. On the other hand, I know the browsers team has been looking forward to this.

> The GC code (I assume) has the necessary code to list all the CIDs under pinset and mfs, so maybe it is not overly complicated to add the revert of those blocks as an extra step for the revert process?

Yeah, we'd need to use the GC coloring code to get all "live" blocks.

ShadowJonathan:
@Stebalien Any progress on this? I came to look up progress on this change from https://github.com/ipfs/go-ipfs/issues/6815

Stebalien:
js-ipfs moved forward with their migration: https://github.com/ipfs/js-ipfs/pull/3124. The go team doesn't have anyone to work on this at the moment.

hsanjuan:
I've brought the branch up to date and will continue where I left off in the good old days.

ipfs-8-to-9/migration/migration.go
hsanjuan: I'm leaving this one as is.

hsanjuan:
> If we upgrade, add some files with CIDv1, then downgrade, we'll GC these files. It'll still be possible to read the files given that we check both V0 and V1 CIDs on read. The only solution I can think of is to revert by walking the pinset/mfs. If we do that, with https://github.com/ipfs/go-ipfs/blob/master/thirdparty/cidv0v1/blockstore.go, we'll avoid GCing anything we want to keep.

This is not addressed yet, but I don't remember if we wanted to address it because the migration was meant to happen on a big go-ipfs release (with high chances for regressions and downgrades). **This happens when the user downgrades after having added CIDv1-addressed content**, in which case, those "new" blocks would stay referenced by multihash in the datastore (as CIDv0). They could still be read, because the blockstore is wrapped insomething that tries reading CIDv1 but fails over to CIDv0, but if a GC runs at that point, they could be GC'ed.

@Stebalien @aschmahmann is this something we still want to address?

The proposed solution is to:

* Walk all pins and mfs root recursively and collect all those CIDs before revert.
* Make sure that all those are stored as CID when the revert happens.

I'm pretty sure there are corner cases still: i.e. if the user, post migration, adds something with CidV0 to MFS, then adds the same with CidV1 and pins. It is likely that we will be removing the CIDv0 block unless we have an additional check to see if the equivalent to a CIDv1 is in the pin/mfs trees as CIDv0. This will add more memory pressure or we have to opt for a bloom filter check.

hsanjuan:
> I'm pretty sure there are corner cases still: i.e. if the user, post migration, adds something with CidV0 to MFS, then adds the same with CidV1 and pins. It is likely that we will be removing the CIDv0 block unless we have an additional check to see if the equivalent to a CIDv1 is in the pin/mfs trees as CIDv0. This will add more memory pressure or we have to opt for a bloom filter check.

Ok, what we can do is to revert everything from the backup file and all the CIDv1 pins from the pinset and mfs and leave the raw-multihash blocks in place (do not delete anything on revert). That would cover the corner cases of the migration deleting something is should not have deleted on revert, it is simple and the user can choose to GC after revert. The downside is the repo size will potentially grow during revert. Note that we are always talking about CIDv1 stuff (not the default).

ipfs-8-to-9/migration/migration.go
hsanjuan: whatever, let's leave like this.

hsanjuan:
The last commit walks the MFS and pinset trees and ensures all those blocks are reverted to CIDv1 addressing.

Also, we **do not delete any blocks on revert** anymore. There are many corner cases about deleting blocks that were indeed expected to be there after migration because they had both CIDv1 and CIDv0.

The last commit mostly reverts the previous one as it would not be necessary to track which blocks had both CIDv1 and CIDv0 addressing if we do not delete them on revert.

aschmahmann:
@hsanjuan does this now mean that reversions are safe (even in the CIDv1 cases), you'll just end up with more data you'll have to GC?

hsanjuan:
> @hsanjuan does this now mean that reversions are safe (even in the CIDv1 cases), you'll just end up with more data you'll have to GC?

Yes. Should be safe.

hsanjuan:
I have added tests.

@gammazero it seems this wants me to publish the new migration to dist.ipfs.io before it can successfully check that `fs-repo-migration -v` is 12. Either tests are bound to break on branches or bound to break when the new migration is published. What is the right approach?

hsanjuan:
One thing to double-check is whether it is ok to use ColoredSet() from go-ipfs v0.8.0 for the reverts (probably yes because it does not convert CIDs).

hsanjuan:
Afaik tests fail because there were broken in master after #130

gammazero:
The tests appear to be passing now.  Maybe network issue in CI?

hsanjuan:
hm thanks for checking @gammazero !

hsanjuan:
Dependency vendoring is in #131

BigLep:
Blocked until the release (https://github.com/ipfs/go-ipfs/issues/8058)

hsanjuan:
I have been testing the migration using a 2TB badger repository on a AWS EC2 machine, with CPU/16GB RAM and 15TB EBS storage (IO optimized).

This repository is additionally full of CIDv1s (2M+ blocks need migration). This takes about a day or two.

* A badger this big needs to bumped FD limits. That is normal and should be ok on a production setup.
* On a first attempt, everything was going well but the processed suffered an OOM event at 200k . I added pprof debugging and started extracting memory heaps for a second attempt. This is currently at block 1.200M and going fine. My suspect for OOM is badger though, not anything in the migration, given the profiles that it is producing.
* After that, need to verify that a new migration detects no blocks to be migrated, and that a revert also works without issues.

gammazero:
Looks like CI failure due to needing to re-vendor.

BigLep:
2021-07-01 conversation notes:
- nft.storage is the most problematic edge cases
- We could look at our gateways to see how many root CIDs on our gateways or v0 vs. v1
- Potential "cheat" for FlatFS (move files instead of add/delete)

aschmahmann (approved):

PR #84, opened 2019-02-22
Migration 7-8 - Encoding key filenames with base32

AuHau:

PR #96, opened 2020-02-28
Migration 8-9 - Encoding key filenames with base32

hsanjuan:
Continue work from #84 .

More context at https://github.com/ipfs/go-ipfs/pull/6012


ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	nameWithoutPrefix := strings.ToUpper(name[len(keyFilenamePrefix):])
+	data, err := base32.StdEncoding.DecodeString(nameWithoutPrefix)
+
+	if err != nil {
+		return "", err
+	}
+
+	decodedName := string(data[:])
```
hsanjuan: just `string(data)` ?

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,152 @@
...
+
+	return decodedName, nil
+}
+
+func (m Migration) Apply(opts migrate.Options) error {
+	log.Verbose = opts.Verbose
+	log.Log("applying %s repo migration", m.Versions())
+
+	keystoreRoot := filepath.Join(opts.Path, "keystore")
```
hsanjuan: put keystore prefix at the top

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,152 @@
...
+func (m Migration) Reversible() bool {
+	return true
+}
+
+const keyFilenamePrefix = "key_"
+
+func isEncoded(name string) bool {
+	if !strings.HasPrefix(name, keyFilenamePrefix) {
+		return false
+	}
```
hsanjuan: already checked in `decode`.

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,152 @@
...
+		log.VLog("migrating key's filename: ", info.Name())
+		encodedName, err := encode(info.Name())
+		if err != nil {
+			return err
+		}
+
+		os.Rename(
+			filepath.Join(keystoreRoot, info.Name()),
+			filepath.Join(keystoreRoot, encodedName),
+		)
```
hsanjuan: error checking

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,152 @@
...
+func (m Migration) Revert(opts migrate.Options) error {
+	log.Verbose = opts.Verbose
+	log.Log("reverting migration")
+
+	keystoreRoot := filepath.Join(opts.Path, "keystore")
+	fileInfos, err := ioutil.ReadDir(keystoreRoot)
+
+	if err != nil {
+		return err
+	}
```
hsanjuan: seems revert and apply are the same with the encode/decode functions switched.

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,152 @@
...
+			continue
+		}
+
+		if isEncoded(info.Name()) {
+			log.Log("skipping ", info.Name(), " as it is already encoded!")
+			continue
+		}
+
+		log.VLog("migrating key's filename: ", info.Name())
+		encodedName, err := encode(info.Name())
```
hsanjuan: should not encode something already encoded, I guess.

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,152 @@
...
+		decodedName, err := decode(info.Name())
+		if err != nil {
+			return err
+		}
+
+		os.Rename(
+			filepath.Join(keystoreRoot, info.Name()),
+			filepath.Join(keystoreRoot, decodedName),
+		)
+	}
```
hsanjuan: error checking

hsanjuan:
@Stebalien so this is more or less ready:

https://github.com/ipfs/go-ipfs/pull/6955 is ready

The migration is ready and has some tests (which also touch the previous bootstrappers migration).

ipfs-8-to-9/migration/migration.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (m Migration) Reversible() bool {
+	return true
+}
+
+const keyFilenamePrefix = "key_"
+
+const keystoreRoot = "keystore"
+
+func isEncoded(name string) bool {
```
Stebalien: Given that std base32 includes all letters, isn't this migration a no-op for any key named with a single lower-case word? (`main`, `ipns`, etc.).?

If you want to be robust, we should copy keys from an old directory to a new directory, then replace the old directory. However, even that isn't perfect as we might fail between replacing the old directory and updating the repo version.

Really, we need some way to record that a migration is in-progress. But we don't currently have a way to do that.

hsanjuan: Decode (and isEncoded()) will fail if the filename do not start with `key_`. Not 100% bulletproof, but exceptions will be logged.

I think this is an ok assumption to make, mostly because repairing the issue if it happens is easy for the user.

Stebalien: Ah, now I see. Nevermind.

PR #97, opened 2020-03-06
7-to-8: allow Bootstrappers to be nil/non-existing.

hsanjuan:
Users may have set Bootstrappers to null to disable them
in some networks. This should not make migrations fail.


hsanjuan:
(sharness tests do `ipfs config Bootstrap --json null` so setting Bootstrappers to null seems like a valid thing to do).

PR #99, opened 2020-03-17
fix: fix build on 32 bit BSDs

Stebalien:
Can't really think of a better way to do this other than to just hack these vendored deps.

Ironically, the go-ipfs version for the migration that's failing never actually _built_ on these platforms. However, fixing this is cleaner than simply excluding the migration for the platform in question.


Stebalien:
I'm going to merge this without review because it _should_ be fine. However, I'd appreciate a post-merge double-check.

hsanjuan:
lgtm. I just hope it didn't take you too long to figure it out..

Stebalien:
I think I've fixed this bug before...

PR #102, opened 2020-05-22
feat: migration 9 to 10

dirkmc:
Fixes https://github.com/ipfs/go-ipfs/issues/7343

### DIFF
--- a/ipfs-9-to-10/migration/config_conv.go
+++ b/ipfs-9-to-10/migration/config_conv.go
@@ -26,10 +26,14 @@ func convertFile(orig string, new string, enableQuic bool, convBootstrap convArr
 	if err != nil {
 		return err
 	}
+	defer in.Close()
+
 	out, err := os.Create(new)
 	if err != nil {
 		return err
 	}
+	defer out.Close()
+
 	return convert(in, out, enableQuic, convBootstrap, convSwarm)
 }
 
@@ -65,9 +69,12 @@ func convert(in io.Reader, out io.Writer, enableQuic bool, convBootstrap convArr
 	if err != nil {
 		return err
 	}
-	out.Write(fixed)
-	out.Write([]byte("\n"))
-	return nil
+
+	if _, err := out.Write(fixed); err != nil {
+		return err
+	}
+	_, err = out.Write([]byte("\n"))
+	return err
 }
 
 // Remove Experimental.QUIC flag
@@ -77,7 +84,11 @@ func removeExperimentalQuic(confMap map[string]interface{}) (bool, bool) {
 		return false, false
 	}
 	enabledi, ok := confExpi["QUIC"]
-	enabled := enabledi.(bool)
+	if !ok {
+		return false, false
+	}
+
+	enabled, ok := enabledi.(bool)
 
 	delete(confExpi, "QUIC")
 
@@ -88,7 +99,7 @@ func removeExperimentalQuic(confMap map[string]interface{}) (bool, bool) {
 func convertBootstrap(confMap map[string]interface{}, conv convArray) {
 	bootstrapi, _ := confMap["Bootstrap"].([]interface{})
 	if bootstrapi == nil {
-		log.Log("Bootstrap field missing or of the wrong type")
+		log.Log("No Bootstrap field in config, skipping")
 		return
 	}
 	bootstrap := make([]string, len(bootstrapi))
--- a/ipfs-9-to-10/migration/config_conv.go
+++ b/ipfs-9-to-10/migration/config_conv.go
@@ -34,6 +34,11 @@ func convertFile(orig string, new string, enableQuic bool, convBootstrap convArr
 	}
 	defer out.Close()
 
+	// Make sure file has 600 permissions
+	if err := out.Chmod(0600); err != nil {
+		return err
+	}
+
 	return convert(in, out, enableQuic, convBootstrap, convSwarm)
 }
 
--- a/ipfs-9-to-10/migration/config_conv.go
+++ b/ipfs-9-to-10/migration/config_conv.go
@@ -27,15 +27,18 @@ func convertFile(path string, enableQuic bool, convBootstrap convArray, convSwar
 	if err != nil {
 		return err
 	}
-	defer in.Close()
 
 	// Create a temp file to write the output to on success
 	out, err := atomicfile.New(path, 0660)
 	if err != nil {
+		in.Close()
 		return err
 	}
 
 	err = convert(in, out, enableQuic, convBootstrap, convSwarm)
+
+	in.Close()
+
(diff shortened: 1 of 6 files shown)


ipfs-9-to-10/migration/config_conv.go
```diff
@@ -0,0 +1,189 @@
...
+// configuration from one version to another
+type convArray func([]string) []string
+
+// convertFile converts a config file from one version to another
+func convertFile(orig string, new string, enableQuic bool, convBootstrap convArray, convSwarm convArray) error {
+	in, err := os.Open(orig)
+	if err != nil {
+		return err
+	}
+	out, err := os.Create(new)
```
hsanjuan, line 29: missing `defer xx.Close()` ?

hsanjuan: Also, doublecheck that the permissions are correct. If this is the configuration, it should probably be 0600 ? But now that I check, my ipfs config has 0644 :/ Did we mess it up in a previous migration?

ipfs-9-to-10/migration/config_conv.go
```diff
@@ -0,0 +1,189 @@
...
+	} else {
+		convertSwarm(confMap, convSwarm)
+	}
+
+	fixed, err := json.MarshalIndent(confMap, "", "  ")
+	if err != nil {
+		return err
+	}
+	out.Write(fixed)
+	out.Write([]byte("\n"))
```
hsanjuan, line 69: Error checking

ipfs-9-to-10/migration/config_conv.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// Remove Experimental.QUIC flag
+func removeExperimentalQuic(confMap map[string]interface{}) (bool, bool) {
+	confExpi := confMap["Experimental"].(map[string]interface{})
+	if confExpi == nil {
+		return false, false
+	}
+	enabledi, ok := confExpi["QUIC"]
+	enabled := enabledi.(bool)
```
hsanjuan, line 80: Will this crash if someone did  `"QUIC": "true"` (instead of `true`) ? Hopefully IPFS did not allow that mistake in the first place. Anyways, worth `enabled, ok :=` and checking rather than panicking.

ipfs-9-to-10/migration/config_conv.go
```diff
@@ -0,0 +1,189 @@
...
+	// Remove quic addresses
+	res := make([]string, 0, len(swarm))
+	for _, addr := range swarm {
+		if !quicRegexp.MatchString(addr) {
+			res = append(res, addr)
+		}
+	}
+
+	return res
+}
```
hsanjuan, line 189: This will remove QUIC listeners that the user had setup. I guess if we want to not remove that we should keep a marker indicating if we changed anything on ver9to10Swarm or not.

ipfs-9-to-10/migration/migration.go
```diff
@@ -0,0 +1,146 @@
...
+	phase, err := readPhase(phasefile)
+	if err != nil {
+		return fmt.Errorf("reading revert phase: %s", err)
+	}
+
+	for ; phase < 3; phase++ {
+		switch phase {
+		case 0:
+			if err := os.Rename(basepath, v10path); err != nil {
+				return err
```
hsanjuan, line 125: another alternative would be to keep the v9 config file around and just rename that, without the need of un-doing changes unless the v9 file does not exist.

ipfs-9-to-10/migration/config_conv.go
```diff
@@ -0,0 +1,189 @@
...
+	return enabled, ok
+}
+
+// Convert Bootstrap addresses to/from QUIC
+func convertBootstrap(confMap map[string]interface{}, conv convArray) {
+	bootstrapi, _ := confMap["Bootstrap"].([]interface{})
+	if bootstrapi == nil {
+		log.Log("Bootstrap field missing or of the wrong type")
+		return
+	}
```
hsanjuan, line 93: More than one person will have bootstrap set to `null` (private networks etc). Perhaps improve the log message as the field will not be really missing, just empty.

ipfs-9-to-10/migration/migration.go
```diff
@@ -0,0 +1,146 @@
...
+	repo := mfsr.RepoPath(opts.Path)
+
+	log.VLog("  - verifying version is '9'")
+	if err := repo.CheckVersion("9"); err != nil {
+		return err
+	}
+
+	basepath := filepath.Join(opts.Path, "config")
+	v9path := filepath.Join(opts.Path, "config-v9")
+	if err := os.Rename(basepath, v9path); err != nil {
```
Stebalien, line 46: We should:

1. Create a temporary file using `atomicfile` (see how go-ipfs-config does this).
2. Write the new config into that temporary file.
3. Atomically rename it.

Stebalien: That way we're never in an inconsistent state.

ipfs-9-to-10/migration/config_conv.go
```diff
@@ -0,0 +1,205 @@
...
+	defer in.Close()
+
+	out, err := os.Create(new)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	// Make sure file has 600 permissions
+	if err := out.Chmod(0600); err != nil {
```
Stebalien, line 38: Ah, now I see what you mean. If we're _renaming_ the config, we don't need to chmod it (we'll just inherit the original mode).

ipfs-9-to-10/migration/migration.go
```diff
@@ -0,0 +1,120 @@
...
+	phase, err := readPhase(phasefile)
+	if err != nil {
+		return fmt.Errorf("reading revert phase: %s", err)
+	}
+
+	defer os.Remove(phasefile)
+	for ; phase < 2; phase++ {
+		switch phase {
+		case 0:
+			if err := convertFile(path, false, ver10to9Bootstrap, ver10to9Swarm); err != nil {
```
Stebalien, line 103: We don't really need to do anything in the reverse migration. The QUIC addresses will be ignored if we don't support them.

Can we even correctly implement the reverse migration?

ipfs-9-to-10/migration/config_conv.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	// Convert bootstrap config
+	convertBootstrap(confMap, convBootstrap)
+
+	// Convert swarm config
+	if enableQuic {
+		// When enabling quic
+		// - Remove experimental option from config
+		confEnabled, ok := removeExperimentalQuic(confMap)
```
Stebalien, line 71: I'd leave this.

PR #104, opened 2020-06-09
fix: correctly check for existing quic addresses

Stebalien:
* And check a some other edge cases.
* And test everything.

fixes https://github.com/ipfs/go-ipfs/issues/7444

----

#


Stebalien:
Ok, all the important tests are passing. I'm not sure why the others are failing.

Issue #98, opened 2020-03-13
Plan proposal for future of fs-repo-migrations


hsanjuan:
## State of the art

`fs-repo-migrations` is a binary that knows how to migrate repository versions since version 1 to soon-to-be version 9. 

Each individual migration is a go-module (and a go executable) but they are all wired together in `fs-repo-migrations`, which is published on dist.ipfs.io and consumed by `ipfs` and `ipfs-update` (in very similar ways) when a migration is necessary (either by http-download or by ipfs-download).

Each migration module needs to be built offline, and they use vendored dependencies. Due to the long history of migrations, these includes library copy-pastes added as extra modules, gxed-vendored dependencies added as submodules and go-mod vendored dependencies.

Migrations are mostly tested with single, non vendored sharness tests, with some deps depending on iptb.

## Current problems

* The `fs-repo-migrations` binary is now 54M (to version 7), soon to grow
* There are some libraries (opencensus tracing related) that only allow to be `init()`ed ONCE. Multiple subpackages may vendor different versions and then fail badly. This needs patching across all migrations to prevent init() from happening or to align the libraries.
* Things like `/net/trace` (if I remember well) do not allow multiple versions and fail. This requires alignment between all submodules.
* Resolving the above involves touching very legacy code and potentially breaking very old migrations. It is horribly annoying, any new contributor will need a core dev to give a hand if this happens.
* It has been discussed to deprecate old migrations (#93) but apparently we do not want that.

## Proposal

We should make it easy for anyone to drop in and write a migration. Migration code should be fully isolated and testable in isolation. We should prevent having to touch any of the old migrations, or to having to adjust test things which affect all migrations and end up becoming a time sink on the developer.

Thus:

1. `fs-repo-migrations` binary is deprecated and we stop using it. We build migration-specific binaries for all migrations.
2. The migration specific binaries are produced in isolation and fully, correctly vendored, without the implementer having to fix retroactively older migrations when a dependency issue comes up. This binaries are in the form `ipfs-version-to-version`. Tests for new migrations (sharness or not) are added in the submodule directly for new migrations
3. Binaries are built and published separately in `go-migrations.ipfs.io` (using the same mechanism as dist.ipfs.io).
[...]

PR #112, opened 2020-11-04
Migrate ipfs v0.7.0 to v0.8.0

gammazero:
- Convert from mdag pinning to datastore pinning


go.mod
```diff
@@ -2,4 +2,16 @@ module github.com/ipfs/fs-repo-migrations
...
+	github.com/ipfs/go-ds-flatfs v0.4.5
+	github.com/ipfs/go-ds-leveldb v0.4.2
+	github.com/ipfs/go-filestore v1.0.0
+	github.com/ipfs/go-ipfs-blockstore v1.0.1
+	github.com/ipfs/go-ipfs-exchange-offline v0.0.1
+	github.com/ipfs/go-ipfs-pinner v0.0.5-0.20201029233447-cad837873ee6
+	github.com/ipfs/go-ipld-format v0.2.0
+	github.com/ipfs/go-merkledag v0.3.2
+	golang.org/x/net v0.0.0-20200114155413-6afb5195e5aa
+)
```
aschmahmann, line 17: @Stebalien WDYT is this fine for now and then we move to implementing https://github.com/ipfs/fs-repo-migrations/issues/98, or should we just do that now?

aschmahmann: Note: there might be alternatives to #98 such as vendoring the dependencies and giving them all migration-specific names so we can keep everything in one binary, but I'm not sure if they'd be any better.

Stebalien: We can't have unvendored deps in this repo but we could try using go mod vendoring.

aschmahmann: > The go mod vendor command constructs a directory named vendor in the main module's root directory that contains copies of all packages needed to support builds and tests of packages in the main module.

Is the suggestion to make new packages + go.mod's in each migration and then to try and vendor, or to copy the approach from the migrations before we were using gx.

ipfs-10-to-11/go.mod
```diff
@@ -0,0 +1,32 @@
...
+
+require (
+	github.com/ipfs/fs-repo-migrations v1.6.4
+	github.com/ipfs/go-blockservice v0.1.2
+	github.com/ipfs/go-cid v0.0.7 // indirect
+	github.com/ipfs/go-datastore v0.4.5
+	github.com/ipfs/go-ds-flatfs v0.4.5
+	github.com/ipfs/go-ds-leveldb v0.4.2
+	github.com/ipfs/go-filestore v1.0.0
+	github.com/ipfs/go-ipfs-blockstore v1.0.0
```
aschmahmann, line 13: should use the pre v1 filestore + blockstore (same versions as in go-ipfs).

ipfs-10-to-11/migration/migration.go
```diff
@@ -0,0 +1,184 @@
...
+			Prefix:    datastore.NewKey("/blocks"),
+			Datastore: fds,
+		},
+		{
+			Prefix:    datastore.NewKey("/"),
+			Datastore: ldb,
+		},
+	}))
+	var dstore datastore.Batching
+	dstore = &batchWrap{mdb}
```
aschmahmann, line 127: This is specific to flatfs + leveldb, but people could be using anything. Instead use `go-ipfs/repo/fsrepo` with `repo, err := fsrepo.Open(repopath)` and `dstore = repo.Datastore()`

Stebalien:
I've manually vendored the deps.

gammazero:
One of the problems with using go modules for vendoring the dependencies of `ipfs-10-to-11`, and making `ipfs-10-to-11` into a module, is that building fs-repo-migrations requires the the `github.com/ipfs/fs-repo-migrations/ipfs-10-to-11` module.  Unfortunately, `ipfs-10-to-11` requires the `github.com/ipfs/fs-repo-migrations`.  This creates a circular dependency

I think the correct solution to this is presented in #98 which gets rid of the monolithic fs-repo-migrations binary.  As a temporary solution `ipfs-10-to-11` can depend on a previous version of `fs-repo-migrations` which had no changes to the packages imported, and did not depend on `ipfs-10-t0-11`.  This does not seem like a sustainable way of doing things, particularly if manual vendoring is required. (huge thanks to @Stebalien for doing that in this PR)

ipfs-10-to-11/migration/migration.go
gammazero: Done in [726f3c7](https://github.com/ipfs/fs-repo-migrations/pull/112/commits/726f3c787f63031cac157e40dfc4b653aa01f2a0)

gammazero:
Note: Needed to update go version in top-level go.mod to `go 1.14` to fix "duplicate method" error when compiling go-ipfs for the 10-to-11 migration.  That error is caused by using the go 1.13 spec specified in go.mod.  See [here](https://dave.cheney.net/tag/interfaces) for details.

ipfs-10-to-11/go.mod
gammazero: Done in [795e3b0](https://github.com/ipfs/fs-repo-migrations/pull/112/commits/795e3b0f85f0e01e10662ff64a050b533dd21724)

PR #117, opened 2021-01-13
Fix error caused by double-locking repo

### LINKED ISSUE #115: Lockfile detection failure on Debian
Using the last snapshot of source code, I compiled the tool on Debian Buster x86. But it is now impossible to migrate the repository because the locking engine reports the lockfile size is not zero. This issue happens also with the LevelDB datastore selected for the repository. So the only working one is BadgerDS but I am not able to use it on such an old machine. Do you know the possible workaround?

gammazero:
This fixes #115.

The repo was being double-locked, once by the migration and once by opening the repo.  However, on some platforms this double locking was not detected due to the combination of different locking implementations, and the defect was not detected by tests.

### DIFF
--- a/ipfs-10-to-11/migration/migration.go
+++ b/ipfs-10-to-11/migration/migration.go
@@ -18,7 +18,6 @@ import (
 	"github.com/ipfs/fs-repo-migrations/ipfs-10-to-11/_vendor/github.com/ipfs/go-merkledag"
 
 	migrate "github.com/ipfs/fs-repo-migrations/go-migrate"
-	lock "github.com/ipfs/fs-repo-migrations/ipfs-1-to-2/repolock"
 	mfsr "github.com/ipfs/fs-repo-migrations/mfsr"
 	log "github.com/ipfs/fs-repo-migrations/stump"
 )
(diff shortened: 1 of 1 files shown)


aschmahmann (approved):
LGTM. Although it might be a good idea to pull `fsrepo.Open` out of the `revertPins`/`transferPins` functions and pass in the repo so that it's obvious we have the repo lock when we do `mfsr.RepoPath(opts.Path).WriteVersion`

ipfs-10-to-11/migration/migration.go
```diff
@@ -49,7 +49,18 @@ func (m Migration) Apply(opts migrate.Options) error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	if err = transferPins(ctx, opts.Path); err != nil {
+	if !fsrepo.IsInitialized(opts.Path) {
+		return fmt.Errorf("ipfs repo %q not initialized", opts.Path)
+	}
```
aschmahmann, line 54: Why is this only used for Apply and not for Revert?

gammazero: Following example from previous migrations.

Done in both places now.

aschmahmann (approved):

Issue #122, opened 2021-02-25
10-to-11 migration fails with s3ds plugin


DeedleFake:
I have a custom compiled IPFS image using `v0.8.0` with `v0.7.0` of the `go-ds-s3` plugin bundled in, as per the instructions in `go-ds-s3`'s README. When I try to run the new version after upgrading from `v0.7.0`, I get the following log output:

```
ipfs version 0.8.0
Found IPFS fs-repo at /data/ipfs
Initializing daemon...
go-ipfs version: 0.8.0-ce693d7
Repo version: 11
System version: amd64/linux
Golang version: go1.14.4
Found outdated fs-repo, migrations need to be run.
  => Looking for suitable fs-repo-migrations binary.
  => None found, downloading.
  => Running: /tmp/go-ipfs-migrate889866779/fs-repo-migrations -to 11 -y
Found fs-repo version 10 at /data/ipfs
===> Running migration 10 to 11...
applying 10-to-11 repo migration
  - opening datastore at "/data/ipfs"
ipfs migration:  migration 10 to 11 failed: cannot open datastore: unknown datastore type: s3ds
  => Failed: /tmp/go-ipfs-migrate889866779/fs-repo-migrations -to 11 -y
The migrations of fs-repo failed:
  migration failed: exit status 1
If you think this is a bug, please file an issue and include this whole log output.
  https://github.com/ipfs/fs-repo-migrations
```

The only thing that I was able to find that was similar was ipfs/go-ipfs#6416, but it was closed without being resolved, presumably because the original author disappeared. I tried starting up a completely unmodified node via the `ipfs/go-ipfs` image on Docker Hub, but that had the exact same problem. How am I supposed to run the migration if the migration tool doesn't support the backend that I've got my data stored in?

DeedleFake:
I've managed to get my node back up and running by rolling back to `v0.7.0` with `v0.6.0` of the plugin, but I'd definitely like to know how to update, if it's at all possible.

Stebalien:
The repo migration tool will load any plugins present in your repo (or it should), but won't be able to use plugins built into go-ipfs itself. The simplest solution is to build a copy of the s3 plugin and drop it into your repo.

Otherwise, you could try building the migration tool with s3 support. If this doesn't work, I can try walking you through that process.

godcong:
@Stebalien 
I also failed to migrate using badger/v2

===> Running migration 10 to 11...
applying 10-to-11 repo migration
[...]

PR #124, opened 2021-03-11
Build the S3 plugin into the ipfs-10-to-11 migration

gammazero:
This addresses issue #122. It allows the ipfs-10-to-11 migration to be run on a s3 repo without having to build the S3 plugin for the migration.


gammazero:
Closing this in favor of only supporting it only in the new migrations implementation.

PR #126, opened 2021-03-20
Add script to build migrations with datastore plugin

gammazero:
- Update documentation

### DIFF
--- a/build-plugin.sh
+++ b/build-plugin.sh
@@ -13,7 +13,7 @@
 set -eou pipefail
 
 function usage() {
-    echo "usage: $0 plugin_repo x-to-y ...">&2
+    echo "usage: $0 plugin_repo[@<version_or_hash>] x-to-y ...">&2
     echo "example: $0 github.com/ipfs/go-ds-s3 10-to-11" >&2
 }
 
@@ -25,6 +25,13 @@ if [ $# -lt 2 ]; then
 fi
 
 plugin_repo="$1"
+if [[ "$plugin_repo" == *"@"* ]]; then
+    plugin_version="$(echo $plugin_repo | cut -d '@' -f 2)"
+    plugin_repo="$(echo $plugin_repo | cut -d '@' -f 1)"
+else
+    plugin_version=latest
+fi
+echo "plugin version: $plugin_version"
 plugin_name="$(echo $plugin_repo | rev | cut -d '-' -f 1 | rev)"
 ds_name="${plugin_name}ds"
 
@@ -73,7 +80,7 @@ function clone_ipfs() {
 function bundle_ipfs_plugin() {
     echo "===> Building go-ipfs with datastore plugin $plugin_name"
     pushd "$BUILD_GOIPFS"
-    go get "${plugin_repo}@latest"
+    go get "${plugin_repo}@${plugin_version}"
     popd
 
     echo "$ds_name ${plugin_repo}/plugin 0" >> "${BUILD_GOIPFS}/plugin/loader/preload_list"
--- a/build-plugin.sh
+++ b/build-plugin.sh
@@ -1,0 +1,118 @@
+#!/bin/bash
+#
+# Building migrations with datastore plugin
+#
+# This script builds migrations with a datastore plugin.  Specify the datastore
+# plugin repo and one or more migrations to build.  Each migration binary is
+# built in its module subdirectory.  Run a migration binary directly, or copy
+# it into a directory in PATH to be run by ipfs-update or fs-repo-migrations.
+#
+# Example:
+# ./build-plugin.sh github.com/ipfs/go-ds-s3 10-to-11 11-to-12
+#
+set -eou pipefail
+
+function usage() {
+    echo "usage: $0 plugin_repo x-to-y ...">&2
+    echo "example: $0 github.com/ipfs/go-ds-s3 10-to-11" >&2
+}
+
+if [ $# -lt 2 ]; then
+    echo "too few arguments" >&2
+    echo >&2
+    usage
+    exit 1
+fi
+
+plugin_repo="$1"
+plugin_name="$(echo $plugin_repo | rev | cut -d '-' -f 1 | rev)"
+ds_name="${plugin_name}ds"
+
+BUILD_DIR="$(mktemp -d --suffix=migration_build)"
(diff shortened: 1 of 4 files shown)


Stebalien:
@aschmahmann status?

aschmahmann:
@gammazero do you mind rebasing this on master and then we can do a walkthrough on how this works and make sure the script is mergeable?

gammazero:
I needed to update the go-ipfs dependency for `fs-repo-10-to-11` because the plugin build script gets the go-ipfs dependency from the migration's `go.mod` file and the version that `fs-repo-10-to-11` previously had was not found in github.

aschmahmann:
@gammazero hmmm... I see we accidentally did a force push https://github.com/ipfs/go-ipfs/pull/7857#event-4514999999. I restored the branch but we should probably update anyway.

If the script works now that I've restored the old branch then let's drop the update from this PR and just do it in another one ideally once the next go-ipfs release is out. Otherwise, we can make a PR that just updates the deps now and have this PR build off of that one.

I want this PR to be two files instead of 700 😄

gammazero:
@aschmahmann After restoring the old branch, the script works.  Now the PR has only one change.

build-plugin.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
+#
+# Building migrations with datastore plugin
```
aschmahmann, line 3: Is any of this datastore specific, or just arbitrary plugins (e.g. could IPLD plugins work too?)

build-plugin.sh
```diff
@@ -0,0 +1,118 @@
...
+# it into a directory in PATH to be run by ipfs-update or fs-repo-migrations.
+#
+# Example:
+# ./build-plugin.sh github.com/ipfs/go-ds-s3 10-to-11 11-to-12
+#
+set -eou pipefail
+
+function usage() {
+    echo "usage: $0 plugin_repo x-to-y ...">&2
+    echo "example: $0 github.com/ipfs/go-ds-s3 10-to-11" >&2
```
aschmahmann, line 17: Two feature requests I'd add here, that can happen after we merge this initial one but would be nice to have are:
- Being able to choose the version of the repo not just the repo itself (e.g. I might need to build an old migration against an old version)
- Being able to add multiple plugins to a given binary

gammazero: - A user can now specify the version of the migration by following it with `@<version_or_hash>`
- The script builds a single migration, with one or more plugins.

Example:
```sh
./build-plugin.sh 10-to-11 github.com/ipfs/go-ds-s3 github.com/ipfs/go-ds-swift@v0.1.0
```

aschmahmann: Awesome, looking forward to giving this a spin.

build-plugin.sh
```diff
@@ -0,0 +1,128 @@
...
+    local plugin_repo="$1"
+    echo "===> Bundling plugin $plugin_repo info go-ipfs for migration"
+    local plugin_version=latest
+    if [[ "$plugin_repo" == *"@"* ]]; then
+        plugin_version="$(echo $plugin_repo | cut -d '@' -f 2)"
+        plugin_repo="$(echo $plugin_repo | cut -d '@' -f 1)"
+    fi
+    echo "plugin version: $plugin_version"
+    local plugin_name="$(echo $plugin_repo | rev | cut -d '-' -f 1 | rev)"
+    local ds_name="${plugin_name}ds"
```
aschmahmann, line 82: I'm pretty sure that the name of the plugin as added to the preload_list is only used for populating the import directives in the code generation script https://github.com/ipfs/go-ipfs/blob/master/plugin/loader/preload.sh.

It'd be nice if the user could set this in case they somehow end up with a conflict, but I don't think it should matter/come up for almost anyone.

build-plugin.sh
```diff
@@ -0,0 +1,128 @@
...
+    
+    pushd "$BUILD_GOIPFS"
+    go get "${plugin_repo}@${plugin_version}"
+    popd
+    echo "$ds_name ${plugin_repo}/plugin 0" >> "${BUILD_GOIPFS}/plugin/loader/preload_list"
```
aschmahmann, line 87: > ${plugin_repo}/plugin 0

This part is incorrect/needs to be configured by the user. The actual subpackage that the plugin is in is totally configurable by the user and so should be something passed in by the user. This doesn't seem like it's too bad since we can just rip it off of the URL that's passed in.

The `0` part is hopefully fine for almost everyone, but technically this can either be any number or '*' and the user should be able to configure it. This is maybe more annoying to pass in, if we needed to default to something maybe choose `*` since it's what we use in our preload examples https://github.com/ipfs/go-ipfs/blob/0ae9b2b9034a57e02df419fe72f705f5a5b3b9c7/plugin/loader/preload_list.

aschmahmann (changes requested):
Does the job for the repos tested, but makes some assumptions about the plugins system we should spell out.

Also, let's add a note in the README that this script exists and might be very useful to people who are running into problems migrating their go-ipfs nodes that use plugins.

build-plugin.sh
gammazero: Since this needs to be specified for each plugin that is build, I think it makes for a much less messy command line if the user is asked interactively about which plugin to load. Now, for each plugin repo build, the user gets a prompt that looks like:
```sh
For github.com/ipfs/go-ds-s3, load 'plugin *' [y/n]?
```
If they answer `n`, then they are asked to for a different value.

The prompting can be avoided by using the `-y` flag as the first argument to the scipt.

build-plugin.sh
gammazero: I assumed this was only for datastore plugins, but I guess it could be built with any.  This is just for building a filestore repo migration, so does it even make sense to support _any_ plugin?  I do not know.

Do I need to add another argument, or prompt,  for the name of the plugin?

build-plugin.sh
gammazero: Done.

When a plugin name conflict is detected, suggest a new name by appending a "1" to the current name.  If running interactive, then ask the user to use the new name or provide one.  If running non-interactive, then keep appending "1" until no conflict.

gammazero:
@aschmahmann Added section to README to describe `build-plugin.sh` script and how to use it.  Script also displays help when given `-h` flag.

aschmahmann (approved):

PR #127, opened 2021-03-22
Rename migrations ipfs-x-to-y to fs-repo-x-to-y

gammazero:


fs-repo-migrations/main.go
```diff
@@ -82,26 +98,45 @@ func main() {
...
 	}
 }
+
+func latestRepoMigration(fetcher migrations.Fetcher) (int, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	var latestMigration int
+	var err error
+	for i := currentVersion; err == nil; i++ {
```
aschmahmann: This is fine for now, but let's file an issue and drop a TODO to remove `currentVersion`. Right now it just saves us a few HTTP queries so why not, but it's also totally doable to just grab the list of all `fs-repo-*-to-*` and calculate it at once without multiple requests.

gammazero: Done.  Added TODO.

PR #130, opened 2021-04-09
Do not ignore arguments

gammazero:
This fixes issue #11

After fix:
```
> fs-repo-migrations/fs-repo-migrations sdff sdfsf sdfsdf
unrecognized arguments
Usage of fs-repo-migrations/fs-repo-migrations:
  -distpath string
        specify the distributions build to use
  -revert-ok
        allow running migrations backward
  -to string
        repo version to upgrade to, or "latest" for latest repo version (default "latest")
  -v    print latest migration available and exit
  -y    answer yes to all prompts
```

### DIFF
--- a/fs-repo-migrations/main.go
+++ b/fs-repo-migrations/main.go
@@ -46,6 +46,12 @@ func main() {
 	yes := flag.Bool("y", false, "answer yes to all prompts")
 	flag.Parse()
 
+	if flag.NArg() != 0 {
+		fmt.Fprintln(os.Stderr, "unrecognized arguments")
+		flag.Usage()
+		os.Exit(1)
+	}
+
 	fetcher := createFetcher(*distPath)
 
 	var (


Stebalien (approved):

hsanjuan:
I think this broke tests in master @gammazero @Stebalien

gammazero:
@hsanjuan That is surprising, since should only complain about arguments that should not be present anyway.  Running `make test` in master branch is passing for me. What tests appear broken?

One thing to be careful of is that some of the tests do not put the test bin directory first in `PATH`, so if you have existing versions of certain binaries that already in `PATH`, it is these versions that are run and not the test versions.

hsanjuan:
The tests did not pass for the commit above, and since merge they do not pass in master, so something broke, perhaps it is not related to this PR but this was still merged with broken CI.

PR #140, opened 2021-11-18
Test via Github actions.

hsanjuan:
And remove travis configuration.


aschmahmann (approved):
It runs so LGTM, thanks 🙏. However, I can't see what version of Go it's using/testing against. It'd be great to see it working against at least Go 1.16 and ideally Go 1.17 as well.

If that's too annoying we can do this in a later PR

hsanjuan:
Let's try

galargh:
I don't think it should be blocking this from being merged but maybe we could also setup unified CI for this repo? I could look into it next week.

aschmahmann:
Unfortunately, this is going to be annoying. We have no guarantees that any of the binaries will be forwards compatible with new versions of Go (this is primarily due to quic-go's reliance on unsafely targeting Go's TLS libraries, although there could be others). Ideally migrations should be written such that we don't need to take a dependency on any of those libraries, but I think that'll take a bit of refactoring such that go-ipfs datastore plugins don't require importing go-ipfs (maybe the Go 1.17 module pruning changes will help us out here).

This makes unified CI not really a great candidate here (some things we'll probably want to copy, but others likely not). We also will be having submodules here which may not work nicely with the unified CI release workflows (not sure though).

We may end up needing to have a mapping of migration versions to versions of Go (e.g. everything up until 10-11 gets 1.15, 10-11 gets 1.15+1.16, 11-12 gets 1.16+1.17) and running them independently.

At some point Actions will probably stop supporting certain versions of Go and that's fine we can drop support, it's pretty rare (although it does occasionally happen) that someone will report a bug in an old migration.

galargh:
> We also will be having submodules here which may not work nicely with the unified CI release workflows (not sure though).

Submodules, if I understand it correctly, are not a problem because we use https://github.com/protocol/multiple-go-modules with all relevant go commands.

> We may end up needing to have a mapping of migration versions to versions of Go (e.g. everything up until 10-11 gets 1.15, 10-11 gets 1.15+1.16, 11-12 gets 1.16+1.17) and running them independently.

This sounds much more tricky and I feel it might be out of scope for unified CI - my intuition tells me it might be special enough of a case to just tackle it here rather than generalising.

> If handling this correctly is too painful right now, we can merge this PR targeting Go 1.16 and then @galargh can resolve this later. At least then we'd have some CI running here.

I strongly agree with this :) 


This is the first failure I can see from a run with Go 1.16:
```
no pre-existing ipfs installation found
fetching go-ipfs version v0.7.0
ERROR: install failed: failed to get ipfs binary: stream error: stream ID 1; INTERNAL_ERROR; received from peer 
install failed, reverting changes...

not ok 2 - 'ipfs-update install' works for v0.7.0
```

aschmahmann:
@galargh it seems like that might've just been a temporary CI issue, things are passing now. As discussed we'll merge and then figure out the CI testing situation across multiple Go versions later.

aschmahmann:
Thanks @hsanjuan 🙏

hsanjuan:
They are not passing?

PR #143, opened 2021-12-10
Use the current path to go-migrate

gammazero:


fs-repo-11-to-12/go.mod
```diff
@@ -4,12 +4,11 @@ go 1.15
 
 require (
 	github.com/hsanjuan/ipfs-lite v1.1.19
-	github.com/ipfs/fs-repo-migrations v1.7.1
```
aschmahmann, line 7: Any reason we're not using a released version here?

aschmahmann (approved):
LGTM, but left a question

fs-repo-11-to-12/go.mod
```diff
@@ -4,12 +4,11 @@ go 1.15
...
+	github.com/ipfs/fs-repo-migrations/tools v0.0.0-20210323144402-297a63449538
```
aschmahmann, line 7: Any reason we're not using a released version here?

aschmahmann: I'm not even sure this is in master, might just be on your local machine from when we did this last time

hsanjuan:
This has failed for something stupid, I'm merging and rebasing #145 on top, and we take those tests as proper signal.

fs-repo-11-to-12/go.mod
gammazero: I cannot use a version tag, as the module in not in the root of the repo.
```
go get: module github.com/ipfs/fs-repo-migrations@v2.0.2 found (v2.0.2+incompatible), but does not contain package github.com/ipfs/fs-repo-migrations/tools
```

If I update to latest, that gives me:
```
github.com/ipfs/fs-repo-migrations/tools v0.0.0-20211210161703-af90efe17fab
```

Which is the v2.0.2 commit, so I will change to that.

PR #145, opened 2021-12-10
11 to 12: Improvements to migration

### LINKED ISSUE #144: 11-to-12: Raw multihash migration
I propose:

* Increasing the number of revert-workers to 4, as there seems to be no reason anymore to keep it in 1 (they have no state that they need to coordinate on anymore)
* Requiring a patched version of go-ds-badger where the gc cycle interval is set to 2 minutes. Otherwise we will not reclaim any space from deleted blocks as we migrate, which may make migrations fail on setups with little free space available on their disks. Knowing the amount of trash badger writes, this may be really shitty (i.e. badger datastore size rising way beyond 2x for datastore full of CIDv1s).

hsanjuan:
Fixes #144

### DIFF
--- a/fs-repo-11-to-12/go.mod
+++ b/fs-repo-11-to-12/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/ipfs/fs-repo-migrations/tools v0.0.0-20211209222258-754a2dcb82ea
 	github.com/ipfs/go-cid v0.0.7
 	github.com/ipfs/go-datastore v0.4.5
+	github.com/ipfs/go-ds-badger v0.2.7-0.20211210151007-a2805355dcf5 // indirect
 	github.com/ipfs/go-filestore v1.0.0
 	github.com/ipfs/go-ipfs v0.8.0
 	github.com/ipfs/go-ipfs-ds-help v1.0.0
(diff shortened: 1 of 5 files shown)


hsanjuan:
This is already rebased on master

fs-repo-11-to-12/go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/ipfs/fs-repo-migrations/tools v0.0.0-20211209222258-754a2dcb82ea
 	github.com/ipfs/go-cid v0.0.7
 	github.com/ipfs/go-datastore v0.4.5
+	github.com/ipfs/go-ds-badger v0.2.7-0.20211210151007-a2805355dcf5 // indirect
```
aschmahmann, line 10: The reliance on a non-tagged commit shouldn't hurt us here since we're using vendoring, right?

fs-repo-11-to-12/migration/swapper.go
```diff
@@ -67,12 +67,7 @@ func (cswap *CidSwapper) Revert(unswapCh <-chan Swap) (uint64, error) {
 	swapWorkerFunc := func() (uint64, uint64) {
 		return cswap.unswapWorker(unswapCh)
 	}
-	// We only run 1 worker for revert. Migrations
-	// many-cid-to-one-multihash mappings, but reverts can have the
-	// opposite. The unswapWorker keeps a cache to handle that, but
-	// this only works with a single worker. Otherwise we'd need
-	// complex syncing, or delayed removal (increased datastore size).
-	return cswap.runWorkers(1, swapWorkerFunc)
+	return cswap.runWorkers(NWorkers, swapWorkerFunc)
```
aschmahmann, line 70: From @hsanjuan: The logic for why we only needed 1 worker is not true anymore. Before we kept track of whether a reverted block had a v0 reference so we did not delete the v0... but now we don't delete anything on revert.

aschmahmann (approved):
LGTM

PR #149, opened 2022-01-14
fs-repo-11-to-12: add env vars for configuring the number of workers and the sync size

aschmahmann:


fs-repo-11-to-12/migration/swapper.go
```diff
@@ -19,6 +21,32 @@ var SyncSize uint64 = 20 * 1024 * 1024 // 20MiB
 // migration.
 var NWorkers int = 4
 
+func init() {
+	workerEnvVar := "IPFS_FS_MIGRATION_11_TO_12_NWORKERS"
+	syncSizeEnvVar := "IPFS_FS_MIGRATION_11_TO_12_SYNC_SIZE"
```
aschmahmann, line 26: Figured long names would prevent conflict and almost nobody should need to set these, but if you have other names in mind that's fine by me.

fs-repo-11-to-12/migration/swapper.go
```diff
@@ -19,6 +21,32 @@ var SyncSize uint64 = 20 * 1024 * 1024 // 20MiB
...
+	if nworkersStr, nworkerInEnv := os.LookupEnv(workerEnvVar); nworkerInEnv {
+		nworkers, err := strconv.Atoi(nworkersStr)
+		if err != nil {
+			panic(err)
+		}
+		if nworkers < 1 {
+			panic("number of workers must be at least 1")
+		}
```
aschmahmann, line 34: We could log an error and fallback to the defaults instead of panicking here, but this seems reasonable to let the user know they did something wrong.

fs-repo-11-to-12/migration/swapper.go
```diff
@@ -19,6 +21,32 @@ var SyncSize uint64 = 20 * 1024 * 1024 // 20MiB
...
+	}
+
+	if syncSizeStr, syncSizeInEnv := os.LookupEnv(syncSizeEnvVar); syncSizeInEnv {
+		syncSize, err := strconv.ParseUint(syncSizeStr, 10, 64)
+		if err != nil {
+			panic(err)
+		}
+		if syncSize < 1 {
+			panic("sync size bytes must be at least 1")
+		}
```
aschmahmann, line 45: Should we set this higher (e.g. 1MB minimum) just so people don't accidentally think 1 == 1MB or something?

Similarly is there any use for setting `0` to mean only syncing at the end of the migration?

sharness/t0030-simple-migration.sh
```diff
@@ -4,6 +4,8 @@ test_description="Simple fs-repo-migrations tests"
 
 . lib/test-lib.sh
 
+latestRepoVersion="12"
```
aschmahmann, line 7: This seems like a wacky change to have to do every time we do a release. Probably we could/should just hard code an immutable distpath so that the repo version is fixed, but at the moment we're just trying to get CI to be green.

fs-repo-11-to-12/migration/swapper.go
hsanjuan: let's rename the env var to sync_size_bytes?

hsanjuan (approved):
I think we should temptatively merge this:

* the env variables work
* it seems badger used storage is not growing as heavily as with the previous settings

I will add a follow-up PR with fixes to multihash parsing etc.

hsanjuan:
And I'm not sure why tests were failing

aschmahmann:
@hsanjuan any learned lessons or recommendations for if/how these variables should be set for us to add to the release notes?

hsanjuan:
@aschmahmann 

> We find that in very large repositories with Badger it is faster to set the number of workers to 1 and increasing the sync size to a larger amount (i.e. 200MB).

PR #150, opened 2022-01-19
11-to-12: Re-work logic to use the backup file for both migrate and revert

hsanjuan:
Before: we Query all keys and write a backup file. Then Query all keys and
    perform the swaps. This may cause that our query keeps getting results for all
    the keys that we are writing while we run it.

After: we perform a Prepare() step where we query all the keys and write the
    backup file. Then we read the backup file to perform the swaps without making
    another query.

This also means that the migration (swapper.Run()) becomes very similar to the
    revert (swapper.Revert()), since both read from the backup file. We are thus
    able to refactor code and re-use logic better.

We have also tuned down the noise when encountering multihashes that are not
    CIDv0s.

### DIFF
--- a/fs-repo-11-to-12/go.mod
+++ b/fs-repo-11-to-12/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/ipfs/fs-repo-migrations/tools v0.0.0-20211209222258-754a2dcb82ea
 	github.com/ipfs/go-cid v0.0.7
 	github.com/ipfs/go-datastore v0.4.5
-	github.com/ipfs/go-ds-badger v0.2.7-0.20211210151007-a2805355dcf5 // indirect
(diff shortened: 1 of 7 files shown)


hsanjuan:
@aschmahmann this is a follow up to #149 . It addresses issues we discussed:

* Do not log errors when encountering blocks that are not CIDv1s nor CIDv0s (potentially raw multihashes with different non-sha256 codecs)
* Use the backup file rather than an ongoing query to perform the migration, similar to how the reverts work.

BigLep:
2022-01-25 discussion: @aschmahmann is going to review and also add some more tests.

aschmahmann:
Talking this through with @hsanjuan this seems to be a pretty reasonable and fairly low risk approach we've done some local testing on. We can do this instead of #151

aschmahmann (approved):

PR #152, opened 2022-02-15
feat: fs-repo-11-to-12 special flatfs handling

aschmahmann:
Trying to add some special casing fast-path code for flatfs.

Some things that are currently missing that may be problematic are:
- We're not using the datastore which means things like disk usage could technically end up out of sync, doesn't seem relevant though since we're not adding or removing data
- Reversions do not leave behind data copies (e.g. a CIDv0 and a CIDv1) which effects both the disk usage and more importantly the discoverability of data. Good news here is that we could use the slow code path for reversion and things should be fine (first let's see if CI catches the bug)
   - CI did catch the bug 🎉, and using the slow path for the reversion seems fine. I hope/suspect reversion of this migration will be very uncommon anyway

### DIFF
--- a/fs-repo-11-to-12/migration/migration_test.go
+++ b/fs-repo-11-to-12/migration/migration_test.go
@@ -17,6 +17,24 @@ const (
 	workingRepo = "repotest_copy"
 )
 
+func TestGenericMigration(t *testing.T) {
+	origSetting := EnableFlatFSFastPath
+	defer func() {
+		EnableFlatFSFastPath = origSetting
+	}()
+	EnableFlatFSFastPath = false
+	testMigrationBase(t)
+}
+
+func TestFlatFSMigration(t *testing.T) {
+	origSetting := EnableFlatFSFastPath
+	defer func() {
+		EnableFlatFSFastPath = origSetting
+	}()
+	EnableFlatFSFastPath = true
+	testMigrationBase(t)
+}
+
 // TestMigration works on an IPFS repository as created by running steps.sh
 // with ipfs v0.8.0 on using the $repotest folder:
 //
@@ -28,7 +46,7 @@ const (
 // added bafybeie4pduk2uwvr5dq36wnbhxspgox7dtqo3fprri4r2wpa7vrej5jqq b
 // added Qmesmmf1EEG1orJb6XdK6DabxexsseJnCfw8pqWgonbkoj c/file3
 // added QmT3zhz9ZZjEpbzWib95EQ5ESUQs4YasrMQwPScpNGLEXZ c
-func TestMigration(t *testing.T) {
+func testMigrationBase(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
--- a/fs-repo-11-to-12/migration/migration_test.go
+++ b/fs-repo-11-to-12/migration/migration_test.go
@@ -194,6 +194,7 @@ func testMigrationBase(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer m.dstore.Close()
 
 	// Check that the CIDv1s that we explicitally pinned or
 	// added to MFS are now retrievable as CIDv1-addressed nodes.
--- a/fs-repo-11-to-12/migration/setup.go
+++ b/fs-repo-11-to-12/migration/setup.go
@@ -2,10 +2,13 @@ package mg11
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
+	"reflect"
+	"unsafe"
 
 	ipfslite "github.com/hsanjuan/ipfs-lite"
 	migrate "github.com/ipfs/fs-repo-migrations/tools/go-migrate"
@@ -18,6 +21,9 @@ import (
 	loader "github.com/ipfs/go-ipfs/plugin/loader"
 	fsrepo "github.com/ipfs/go-ipfs/repo/fsrepo"
 	format "github.com/ipfs/go-ipld-format"
+
+	"github.com/ipfs/go-datastore/mount"
+	flatfs "github.com/ipfs/go-ds-flatfs"
 )
 
 // locks the repo
@@ -81,6 +87,58 @@ func (m *Migration) open(opts migrate.Options) error {
 	}
 	m.dstore = dstore
 	return nil
+}
+
+func getUnexportedField(field reflect.Value) interface{} {
+	return reflect.NewAt(field.Type(), unsafe.Pointer(field.UnsafeAddr())).Elem().Interface()
+}
+
+func IsBasicFlatFSBlockstore(dstore ds.Datastore) (dsPath string, v1 *flatfs.ShardIdV1, err error) {
+	errNotDefault := errors.New("not the default config")
+	defer func() {
+		if err := recover(); err != nil {
+			err = errNotDefault
+		}
+	}()
+
+	mds, ok := dstore.(*mount.Datastore)
+	if !ok {
+		return "", nil, errNotDefault
+	}
+
+	mnts, ok := getUnexportedField(reflect.ValueOf(mds).Elem().FieldByName("mounts")).([]mount.Mount)
+	if !ok {
+		return "", nil, errNotDefault
+	}
+
+	if len(mnts) != 2 {
+		return "", nil, errNotDefault
+	}
+
+	var blkDs ds.Datastore
+	if mnts[0].Prefix.Equal(blocksPrefix) {
+		blkDs = mnts[0].Datastore
+	} else if mnts[1].Prefix.Equal(blocksPrefix) {
+		blkDs = mnts[1].Datastore
+	} else {
+		return "", nil, errNotDefault
+	}
+
+	if reflect.TypeOf(blkDs).String() != "*measure.measure" {
+		return "", nil, errNotDefault
+	}
+
+	fsds, ok := getUnexportedField(reflect.ValueOf(blkDs).Elem().FieldByName("backend")).(*flatfs.Datastore)
+	if !ok {
+		return "", nil, errNotDefault
+	}
+	fsdsPath := reflect.ValueOf(fsds).Elem().FieldByName("path").String()
+
+	shard, err := flatfs.ParseShardFunc(fsds.ShardStr())
+	if err != nil {
+		return "", nil, errNotDefault
+	}
+	return fsdsPath, shard, nil
 }
 
 // Create a file to store the list of migrated CIDs. If it exists, it is
--- a/fs-repo-11-to-12/migration/setup.go
+++ b/fs-repo-11-to-12/migration/setup.go
@@ -99,25 +99,25 @@ func getUnexportedField(field reflect.Value) interface{} {
 }
 
 func IsBasicFlatFSBlockstore(dstore ds.Datastore) (dsPath string, v1 *flatfs.ShardIdV1, err error) {
-	errNotDefault := errors.New("not the default config")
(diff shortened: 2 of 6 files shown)


fs-repo-11-to-12/migration/setup.go
```diff
@@ -83,6 +89,58 @@ func (m *Migration) open(opts migrate.Options) error {
...
+	if !ok {
+		return "", nil, errNotDefault
+	}
+
+	if len(mnts) != 2 {
+		return "", nil, errNotDefault
+	}
+
+	var blkDs ds.Datastore
+	if mnts[0].Prefix.Equal(blocksPrefix) {
```
aschmahmann, line 119: Should I be checking that the other mount is for `/` here?

fs-repo-11-to-12/migration/swapper.go
```diff
@@ -177,10 +180,123 @@ func (cswap *CidSwapper) prepareWorker(resultsCh <-chan query.Result) (uint64, u
...
+		dir := filepath.Join(fsdsPath, fsdsShard.Func()(noslash))
+		file := filepath.Join(dir, noslash+flatfsExtension)
+
+		return dir, file
+	}
+
+	genericSwker := &swapWorker{
+		store:      cswap.Store,
+		syncPrefix: cswap.Prefix,
+	}
```
aschmahmann, line 201: This isn't super pretty, the generic worker is running inside the regular worker because the data is streamed into the workers for the blockstore and filestore together instead of separately, so the worker is either a "generic" or a "generic + flatfs for /blocks" one.

Not sure how much this is worth messing around with, thoughts?

fs-repo-11-to-12/migration/swapper.go
```diff
@@ -177,10 +180,123 @@ func (cswap *CidSwapper) prepareWorker(resultsCh <-chan query.Result) (uint64, u
...
+			log.Error("could not swap %s->%s. Skipping.", sw.Old, sw.New, err.Error())
+			errored++
+			continue
+		}
+		swapped++
+
+		const swapLogThreshold = 10000
+		if swapped%swapLogThreshold == 0 {
+			log.Log("Migration worker has moved %d flatfs files", swapLogThreshold)
+		}
```
aschmahmann, line 261: I wanted to see some progress and it looks like the standard worker also logs per-worker instead of for the system as a whole so I did this here (although flatfs + standard logging are tracked separately).

Thoughts?

aschmahmann (commented):
Some preliminary testing on my local machine indicates this is much faster, so it'd be great if we could land this.

fs-repo-11-to-12/migration/swapper.go
```diff
@@ -177,10 +180,123 @@ func (cswap *CidSwapper) prepareWorker(resultsCh <-chan query.Result) (uint64, u
...
+			log.Error("could not swap %s->%s. Could not find %s even though it was in the backup file %s. Skipping.", sw.Old, sw.New, sw.Old, err.Error())
+			continue
+		}
+
+		if err := os.Mkdir(newDir, 0755); err != nil && !os.IsExist(err) {
+			log.Error("could not swap %s->%s. Skipping.", sw.Old, sw.New, err.Error())
+			continue
+		}
+
+		if err := os.Rename(oldPath, newPath); err != nil {
```
aschmahmann, line 251: Does FlatFS require us to do some cleanup where we go and remove all folders like `1A` that are empty at the end of the migration?

fs-repo-11-to-12/migration/swapper.go
```diff
@@ -177,10 +180,123 @@ func (cswap *CidSwapper) prepareWorker(resultsCh <-chan query.Result) (uint64, u
...
+			}
+			continue
+		}
+
+		_, oldPath := getPath(fsdsPath, sw.Old)
+		newDir, newPath := getPath(fsdsPath, sw.New)
+
+		_, err := os.Stat(oldPath)
+		if err != nil {
+			log.Error("could not swap %s->%s. Could not find %s even though it was in the backup file %s. Skipping.", sw.Old, sw.New, sw.Old, err.Error())
```
aschmahmann, line 242: This mimics the log output from the generic worker, but it ends up making a lot of noise if the migration is stopped and started because a whole lot of data won't be found since it's already been moved.

There's a similar result with the check for CIDs since many of them are multihashes after the stopping and restarting the migration.

I'm wondering if we want to reduce that logging at all, or put it under LogV

aschmahmann:
Did some local benchmarking:

On my HDD (running on Windows)
- 190k FlatFS moves in ~10minutes (Disk usage showed 100% on that drive which is only being used for this migration)
- 16k regular moves in ~11 min (Disk usage was ~90% on the drive)

I suspect SSDs which have better random access will perform better here. There's probably also clever ways to group up renames to maximize mechanical sympathies, but what we have now seems like a pretty good improvement.

---

Looking at this migration running on my HDD (using a single worker) I'm actually a little concerned that users with lots of data might be annoyed at the disk usage during the migration and might ask us to tune it down. If users have non-huge repos and are using SSDs (e.g. most people on laptops and desktops) they should be fine though. WDYT?

hsanjuan:
> * first let's see if CI catches the bug)

:-) I am very proud of my tests

hsanjuan (commented):
This looks ok to me.

We should test it.

My only question is whether this should have a flag to enable (disabled by default), or to disable (enabled by default).

I guess the main danger is that this breaks on some weird mount configuration or something, but seems to take these things into account.

fs-repo-11-to-12/migration/swapper.go
```diff
@@ -177,10 +180,123 @@ func (cswap *CidSwapper) prepareWorker(resultsCh <-chan query.Result) (uint64, u
...
+			continue
+		}
+
+		if err := os.Mkdir(newDir, 0755); err != nil && !os.IsExist(err) {
```
guseggert, line 246: Should we need to preserve dir perms here?

fs-repo-11-to-12/migration/swapper.go
hsanjuan: @aschmahmann can you change this to log the total number of files moved (swapped)? Now it just repeats:

```
Migration worker has moved 10000 flatfs files
Migration worker has moved 10000 flatfs files
Migration worker has moved 10000 flatfs files
Migration worker has moved 10000 flatfs files
Migration worker has moved 10000 flatfs files
Migration worker has moved 10000 flatfs files
Migration worker has moved 10000 flatfs files
```

guseggert (commented):
TBH, pushing this in at the last second makes me slightly uncomfortable because it hasn't been exercised by users w/ the RC. 

The users who benefit from this are the ones with default datastore configs, are there a lot of such users who have huge repos that would benefit from this optimization? What's the "typical" speedup that a typical user would see?

fs-repo-11-to-12/migration/swapper.go
aschmahmann: I grabbed this from https://github.com/ipfs/go-ds-flatfs/blob/40338ecd3b4677a358e1ae1580f7bd324d650db1/flatfs.go#L345. Seems like it's what FlatFS normally does unless I'm mistaken.

fs-repo-11-to-12/migration/setup.go
```diff
@@ -83,6 +89,58 @@ func (m *Migration) open(opts migrate.Options) error {
 	return nil
 }
 
+func getUnexportedField(field reflect.Value) interface{} {
+	return reflect.NewAt(field.Type(), unsafe.Pointer(field.UnsafeAddr())).Elem().Interface()
+}
+
+func IsBasicFlatFSBlockstore(dstore ds.Datastore) (dsPath string, v1 *flatfs.ShardIdV1, err error) {
+	errNotDefault := errors.New("not the default config")
```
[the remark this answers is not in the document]
aschmahmann: It's not quite the default, probably the error here should change, what we want is to constrain to a single FlatFS repo with all your blocks. Mostly we're trying to limit our risk profile here.

aschmahmann:
> The users who benefit from this are the ones with default datastore configs, are there a lot of such users who have huge repos that would benefit from this optimization? What's the "typical" speedup that a typical user would see?

Hard to say. I think most people have the default configs and some people with really big repos have started coming back to FlatFS after trying out Badger and running into problems, so this is a bunch of users.

The larger repo I was testing on locally (although I didn't finish the full migration) was about 80GB with 1.2M CIDv1s and I was seeing a 10x speedup. I'd suspect typical users might see even bigger speedups if they have SSDs, but haven't directly tested that yet.

> TBH, pushing this in at the last second makes me slightly uncomfortable because it hasn't been exercised by users w/ the RC.

True and me too, although I'm not sure what our exercise rate on v0.12.0-rc1 has been despite how long it's been out for. In theory we can just do RC2 and wait a week or so and do the final release, but I'm not sure how much value that brings us.

fs-repo-11-to-12/migration/swapper.go
aschmahmann: Added some logging changes in 7bdf4b4361dff3ae04887299f560eac6489165f8

guseggert (approved):

fs-repo-11-to-12/migration/setup.go
aschmahmann: It doesn't seem necessary to me given that we're checking for prefix equality (e.g. people shouldn't be able to mix things up if they've divided up the blockstore across multiple datastores)

fs-repo-11-to-12/migration/swapper.go
aschmahmann: Passing on messing around with this

fs-repo-11-to-12/migration/swapper.go
aschmahmann: local testing of just creating a default go-ipfs repo and cleaning out all the pins + GCing results in plenty of empty folders so shouldn't be any problems here

fs-repo-11-to-12/migration/swapper.go
hsanjuan: all good ideas

fs-repo-11-to-12/migration/migration_test.go
```diff
@@ -17,6 +17,24 @@ const (
...
+func TestGenericMigration(t *testing.T) {
+	origSetting := EnableFlatFSFastPath
+	defer func() {
+		EnableFlatFSFastPath = origSetting
+	}()
+	EnableFlatFSFastPath = false
+	testMigrationBase(t)
+}
+
+func TestFlatFSMigration(t *testing.T) {
```
aschmahmann, line 29: Unfortunately, running two migration tests results in:

```
    migration_test.go:67: error injecting plugins: already have a datastore named "badgerds"

```

we can either move around the plugin injector code, or just cheat and use unexported global variables here that let the tests work.

aschmahmann: The datastore plugins end up as singletons anyway (https://github.com/ipfs/go-ipfs/blob/5922eab08f22dc815222bcdc3751165b29c82e28/repo/fsrepo/datastores.go#L67) so I just promoted this to be a package wide singleton

guseggert (approved):

aschmahmann:
Did some local testing of env vars just to double check (Go tests aren't too happy about dealing with them) and everything looks good so we're going to hit some buttons and do a release 🎉

fs-repo-11-to-12/migration/setup.go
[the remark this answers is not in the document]
aschmahmann: ah, will fix this text too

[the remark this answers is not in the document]
aschmahmann: done

PR #154, opened 2022-02-23
bump current version to 12

Luflosi:
The currentVersion is outdated, bump it to 12.


BigLep (approved):

BigLep:
We're not going to release new binaries when the version bump is the only change.  This is correct though so we're merging.

BigLep:
We're not going to release new binaries when the version bump is the only change. This is correct though so we're merging.

PR #181, opened 2023-09-11
14: skip unknown types

Jorropo:


fs-repo-14-to-15/migration/migration.go
```diff
@@ -175,7 +175,8 @@ func convert(in io.Reader, out io.Writer) error {
 	if a, ok := confMap["Addresses"]; ok {
 		addresses, ok := a.(map[string]any)
 		if !ok {
-			return fmt.Errorf("invalid type for .Addresses got %T expected json map", a)
+			fmt.Printf("invalid type for .Addresses got %T expected json map; skipping .Addresses\n", a)
+			goto addressesAreNotOk
```
marten-seemann, line 179:
```suggestion
			goto fail
```

fs-repo-14-to-15/migration/migration.go
```diff
@@ -202,20 +204,23 @@ func convert(in io.Reader, out io.Writer) error {
 			addresses[addressToRemove] = newSwarm
 		}
 	}
+addressesAreNotOk:
```
marten-seemann, line 207:
```suggestion
fail:
```

marten-seemann (commented):
Suggestions are not meant to be taken seriously. Instead, consider refactoring to not use `goto`.

fs-repo-14-to-15/migration/migration.go
```diff
@@ -172,65 +172,79 @@ func convert(in io.Reader, out io.Writer) error {
...
-					if !strings.Contains(addr, "/quic") || strings.Contains(addr, "/quic-v1") {
-						newSwarm = append(newSwarm, addr)
+				var newSwarm []interface{}
+				for _, v := range swarm {
+					if addr, ok := v.(string); ok {
+						if !strings.Contains(addr, "/quic") || strings.Contains(addr, "/quic-v1") {
+							newSwarm = append(newSwarm, addr)
+						}
+					} else {
+						newSwarm = append(newSwarm, v)
```
aschmahmann, line 202: I know this was here before, but what is this if not a string? Are we just keeping it so as not to interfere?

fs-repo-14-to-15/migration/migration.go
```diff
@@ -172,65 +172,79 @@ func convert(in io.Reader, out io.Writer) error {
...
 
-			var newSwarm []interface{}
-			for _, v := range swarm {
-				if addr, ok := v.(string); ok {
-					if !strings.Contains(addr, "/quic") || strings.Contains(addr, "/quic-v1") {
-						newSwarm = append(newSwarm, addr)
+				var newSwarm []interface{}
+				for _, v := range swarm {
+					if addr, ok := v.(string); ok {
+						if !strings.Contains(addr, "/quic") || strings.Contains(addr, "/quic-v1") {
```
aschmahmann, line 198: Not sure if we have good options, but it's pretty awkward that we're removing `/quic` here  when we have no idea if there's a substitute `/quic-v1` in place. Should we look for identical matches instead, otherwise we're going to delete the users' data that they may have added to `AppendAnnounce`, etc.

Given the user has to do something here would we be better off:
1. Changing their custom config to `/quic-v1`
2. Deleting `/quic`
3. Letting the user get an error on start and figure it out (error messages in kubo need to be helpful enough for this to be viable, they might already be)

Seems like probably not a ton of people will be impacted, but the support time here might be no fun unless this is taken care of 😢.

fs-repo-14-to-15/migration/migration.go
```diff
@@ -172,65 +172,79 @@ func convert(in io.Reader, out io.Writer) error {
...
-		for _, addressToRemove := range [...]string{"Swarm", "Announce", "AppendAnnounce", "NoAnnounce"} {
-			s, ok := addresses[addressToRemove]
+	if err := func() error {
+		if a, ok := confMap["Addresses"]; ok {
+			addresses, ok := a.(map[string]any)
 			if !ok {
-				continue
+				fmt.Printf("invalid type for .Addresses got %T expected json map; skipping .Addresses\n", a)
+				return nil
 			}
```
aschmahmann, line 181: What kinds of things are we expecting here? Is the idea that we don't really care, it's not something with a valid `/quic` address we can deal with?

aschmahmann (commented):
Haven't run the migration to test yet, but aside from the comment on substitution seems ok.

Some tests on the error cases might be nice if not too painful.

fs-repo-14-to-15/migration/migration.go
Jorropo: mb should be a continue

fs-repo-14-to-15/migration/migration.go
Jorropo: we already had a migration that duplicated quic into quic-v1 addresses, do you want me to copy paste it and rerun it again ?

fs-repo-14-to-15/migration/migration.go
Jorropo: yes

fs-repo-14-to-15/migration/migration.go
aschmahmann: Well don't want to end up with duplicate quic-v1 addresses. Could do more work here to make sure we're not, but a bit concerned about added surface needing testing.

Might be fine leaving things alone given it meant that after quic-v1 addresses were in the config the user added only a `/quic` address.

WDYT?

BigLep:
> Some tests on the error cases might be nice if not too painful.

Per 2023-09-12 conversation: we should get tests so we're not relying on manual tests.  Right now we support a single input/output test.  This should be expanded to support multiple.

fs-repo-14-to-15/migration/migration.go
lidel: +1, it will be safer is we effectively re-run migration and convert all `/quic` to `/quic-v1` and then remove any `/quic-1` duplicates. Way safer than blindly removing `/quic` from user config.

Jorropo:
About testing for `null` I didn't made two test suites since this made the tests bigger than required.
So I just put under different fields in the end.

lidel (approved):
lgtm, i've tested manually the scenario described by Adin and works fine.

I think its ok to ship, we've  spent enough time on this already, but if we end up having to fix this migration one more time :upside_down_face: then we will create a separate test for every edge case.

PR #185, opened 2023-09-20
docs: fix fs-repo link in run.md

SgtPooki:
I am not sure what the right link should be.


lidel (approved):
That is kubo-specific spec, we want to remove those from ipfs/specs, so removed the link here.

PR #186, opened 2023-10-12
fix(14-to-15): ignore Bootstrap empty

Dreamacro:
if kubo startup with `ipfs bootstrap rm all`, the bootstrap list is empty.

current output
```
applying 14-to-15 repo migration
> Upgrading config to new format
error: invalid type for .Bootstrap got <nil> expected json array
```

### DIFF
--- a/fs-repo-14-to-15/migration/migration.go
+++ b/fs-repo-14-to-15/migration/migration.go
@@ -162,7 +162,7 @@ func convert(in io.Reader, out io.Writer) error {
 	}
 
 	// Upgrade bootstrapper QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ from /quic to /quic-v1
-	if b, ok := confMap["Bootstrap"]; ok {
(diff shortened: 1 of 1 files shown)


Jorropo (approved):
Thx, I overlooked that a zero value in a map is different from a missing value.
Now it is obvious it make sense for `encoding/json` to store nils in maps so you can still view the value of the keys.

I'll do a release of the 14 to 15 migration and include the updated CID in Kubo v0.24,

**If everyone else is having this issue please add `"Bootstrap":[]` to your config.** Or update to Kubo v0.24 when it is out.

---

I might add a test later (why not merging yet), I don't have time to release this and upload it to dist.ipfs.tech today.

gammazero:
@lidel Do we sill need to wait for a migration, or can we just merge this fix?

gammazero:
@lidel Do we want to bother making this migration, or should we just close this with with a known work-around.

lidel:
I think this ship sailed, we are closing this  because we don't want  to re-release penultimate migration.

PR #190, opened 2024-08-01
feat: migration from 15 to 16

lidel:
This PR adds config migration that appends `/webrtc-direct` everywhere where UDP `/quic-v1$` listener is already present.

Context: https://github.com/ipfs/kubo/pull/10463

- [x] migration and reverse migration in `fs-repo-15-to-16/migration.go`
- [x] tests in fs-repo-15-to-16/test-e2e
- [x] CI is green


fs-repo-15-to-16/migration/migration.go
lidel, line 1: :information_source:  This PR includes a lot of boilerplate, but the only code that changed is in `fs-repo-15-to-16/migration.go` + tests in `fs-repo-15-to-16/test-e2e`

lidel (commented):
- CI is green, was failing due to 504 from the public gateway.
  - I've added explicit redirect from `dist-ipfs-[io|tech].ipns.dweb.link` to `dist.ipfs.tech` so legacy tests hit the correct infrastructure right away, rather than going via a proxy.
- The code in PR here is 99% copy&paste from previous migration, so not much to review, just a lot of boilerplate.
  - Merging, will now open PR against https://github.com/ipfs/distributions which will produce a CID that I can use for end-to-end testing in Kubo against real public gateway URLs and IPFS routing.

